PR #9, opened 2024-03-29
Implement a tile representation for base edition of the game

### LINKED ISSUE #5: Implement a structure representing tiles and map tiles from base game to it
We need to implement a structure capable of representing tiles from the base game.

All of the tiles from the base game (their road, city, monastery, and field layout + counts of each unique tile layout) should be documented (with images). Their tile layout should be written down in a format compatible with the implemented structure (ideally in a separate sheet or similar file that can later be extracted into a serialized form).

Kaszana102:
Defined structures:
Tiles
Roads
Fields
Cities

All with method rotate
Tests included
connections implemented as "lists"

set of tiles for standard game created

Resolves #5


pkg/tiles/set_of_tiles.go
```diff
@@ -0,0 +1,848 @@
...
+)
+
+func GetStandardTiles() []Tile {
+	var tiles []Tile
+	//from wikipedia from left to right, up to down
+
+	//monastery without roads
+	for i := 0; i < 5; i++ {
+		tiles = append(tiles,
+			Tile{
```
bstrzelecki, line 16: make it into a constant, so it would be easier to create custom tile stacks

pkg/tiles/tile.go
```diff
@@ -0,0 +1,67 @@
...
+	HasShield bool
+	Building  buildings.Bulding
+
+	//dać Building po prostu by skomponować
+
+	//not sure how to include undefied/null?
+	//meeple    Meeple
+}
+
+func (tile *Tile) Rotate(rotations uint) Tile {
```
bstrzelecki, line 22: take self parameter as copy so it wouldn't be necessary to change this method while extending Tile struct

pkg/tiles/connection/connection.go
```diff
@@ -0,0 +1,105 @@
+package connection
+
+type Side int64
+
+const (
```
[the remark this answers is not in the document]
Kaszana102: but why?

Corner position are made for farmers.

pkg/tiles/farm_connection/farm_connection.go
```diff
@@ -0,0 +1,111 @@
+package farm_connection
+
+/**
+First direction is from center, then specyfing which side of edge
+*/
+type FarmSide int64
+
+const (
```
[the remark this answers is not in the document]
Kaszana102: mostly, but the idea is different. 
Connection indicates these directions:
  \  /\   /
    \ | /
<----->
   /  | \
 /   \/   \
8 of them.
According to the article there are only 5 position farmer can be placed (those corners and center).

Farm connection makes distinction between side of the edge of a tile (left/right, bottom/top). For example left side of top edge. It's not the same as corner. 

I would either leave as it is, or make special enum for farmers.

pkg/tiles/cities.go
```diff
@@ -0,0 +1,7 @@
+package tiles
+
+import connection "github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tiles/connection"
+
+type Cities struct {
```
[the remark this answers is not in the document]
Kaszana102: I don't get it

pkg/tiles/buildings/bulding.go
```diff
@@ -0,0 +1,20 @@
+package buildings
+
+type Bulding int64
+
+const (
+	NONE_BULDING Bulding = iota
+	MONASTERY
+)
+
+func (building Bulding) ToString() string {
```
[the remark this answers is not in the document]
Kaszana102: i didn't know how to debug apart from print, and couldn't find it on the internet. It made debugging much easier for me.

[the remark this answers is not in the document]
bstrzelecki: https://github.com/go-delve/delve

Jackenmen (changes requested):
Aside from my review comments, I have a general comment on the variable names in tests - please use more meaningful names to indicate the actual and expected output of the tests. Standard names are `actual` for the actual output that we got from the thing that we're testing and `expected` for the expected output.

pkg/tile_sets/tile_sets.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tiles"
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tiles/tile_templates"
+)
+
+func GetStandardTiles() []tiles.Tile {
+	var tiles []tiles.Tile
+	//from wikipedia from left to right, up to down
+
+	//monastery without roads
+	for i := 0; i < 5; i++ {
```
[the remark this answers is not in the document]
Kaszana102: but this way it doesn't compile 
obraz_2024-04-07_093249056

pkg/tiles/tile.go
```diff
@@ -0,0 +1,65 @@
...
+		t.Fields.Fields = append(t.Fields.Fields, field.Rotate(rotations))
+	}
+
+	//copy parameters
+	t.HasShield = tile.HasShield
+	t.Building = tile.Building
+	return t
+}
+
+func (tile *Tile) ToString() string {
```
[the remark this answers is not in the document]
Kaszana102: but why to we want to pass a copy of it?

pkg/tile_sets/tile_sets.go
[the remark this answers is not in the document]
bstrzelecki: Colon is redundant in this case

pkg/tiles/tile.go
```diff
@@ -0,0 +1,65 @@
...
+
+type Tile struct {
+	Cities
+	Roads
+	Fields
+	HasShield bool
+	Building  buildings.Bulding
+
+	//not sure how to include undefied/null?
+	//meeple    Meeple
```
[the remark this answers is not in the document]
bstrzelecki: I do agree, such fields as meeple or position data shouldn't be part of the Tile struct

pkg/tiles/feature.go
```diff
@@ -0,0 +1,17 @@
...
+const (
+	NONE FeatureType = iota
+	ROADS
+	CITIES
+	FIELDS
+)
+
+type Feature struct {
+	FeatureType FeatureType
+	Connections []connection.Connection
```
Jackenmen, line 16: Hmm, it seems that we may not have the same idea about the design that we want to have - I was expecting `Feature` to be a specific *connection* (so we would get rid of `Connection` struct entirely) and have the following fields:
```
type Feature struct {
    Type FeatureType
    Sides []connection.Side
}
```
That would mean that the `Tile.Features` could contain multiple features of the same type and the functions such as `Cities()` would be generating a slice returning all features whose type matches the type that it's supposed to return.

@bstrzelecki did you want something like what I'm suggesting or does the current implementation match what you wanted already?

Jackenmen (changes requested):

pkg/tiles/feature.go
Kaszana102: But this way we wouldn't we able to differentiate connected and unconnected corners

Jackenmen: Why not?

image
```golang
Features: []Feature{
    {Type: CITY, Sides: []Side{connection.Top}},
    {Type: CITY, Sides: []Side{connection.Right}},
}
```

image
```golang
Features: []Feature{
    {Type: CITY, Sides: []Side{connection.Top, connection.Right}},
}
```

Kaszana102: :<

pkg/tiles/buildings/bulding.go
```diff
@@ -0,0 +1,20 @@
+package buildings
+
+type Bulding int64
+
+const (
+	NoneBuilding Bulding = iota
```
bstrzelecki, line 6: Building in NoneBuilding is redundant and could be simplified to just None

pkg/tiles/connection/side_test.go
```diff
@@ -0,0 +1,156 @@
...
+	if None.Rotate(1) != None {
+		t.Fatalf("got %#v should be %#v after rotation", None.Rotate(1), None)
+	}
+
+	if Side(80).Rotate(1) != None {
+		t.Fatalf("got %#v should be %#v after rotation", Side(80).Rotate(1), None)
+	}
+}
+
+func TestSideToString(t *testing.T) { //nolint:gocyclo // simply testing all states
```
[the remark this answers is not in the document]
Kaszana102: I mean, if it's made and correct why would we?

pkg/tiles/buildings/building_test.go
```diff
@@ -0,0 +1,20 @@
+package buildings
+
+import (
+	"testing"
+)
+
+func TestBuildingToString(t *testing.T) {
```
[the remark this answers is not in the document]
Kaszana102: I mean, if it's made and correct why would we?

pkg/tiles/buildings/bulding.go
Kaszana102: noned

pkg/tiles/feature.go
Kaszana102: changed

pkg/tiles/buildings/building_test.go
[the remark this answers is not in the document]
Kaszana102: removed

pkg/tiles/feature.go
```diff
@@ -0,0 +1,17 @@
+package tiles
+
+import connection "github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tiles/connection"
+
+type FeatureType int64
+
+const (
+	NONE FeatureType = iota
+	ROADS
```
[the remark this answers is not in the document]
Kaszana102: changed

pkg/tilesets/tilesets.go
```diff
@@ -0,0 +1,134 @@
...
+		tiles = append(tiles, tiletemplates.XCrossRoad())
+	}
+
+	// 1 city edge no roads
+	for range 5 {
+		tiles = append(tiles, tiletemplates.SingleCityEdgeNoRoads())
+	}
+
+	// 1 city edge straight road
+	for range 4 {
```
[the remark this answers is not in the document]
Kaszana102: changed

pkg/tiles/tile.go
```diff
@@ -0,0 +1,90 @@
...
+
+	// rotate fields
+	for _, field := range tile.Fields() {
+		t.FieldsAppendConnection(connectionMod.RotateSideArray(field, rotations))
+	}
+
+	// copy parameters
+	t.HasShield = tile.HasShield
+	t.Building = tile.Building
+	return t
```
bstrzelecki, line 89: In this case we need to return tile, since otherwise this function becomes a maintenance burden because modifications to Tile structs needs to be reflected in this function. Currently adding new field to the Tile struct will cause resource leak when calling this function.

pkg/tiles/connection/connection.go
```diff
@@ -0,0 +1,164 @@
...
+		return "ERROR"
+	}
+}
+
+/*
+Rotates side clockwise
+*/
+func (side Side) Rotate(rotations uint) Side { //nolint:gocyclo // splitting into multiple switches would be obscure
+	// limit rotations
+	rotations %= 4
```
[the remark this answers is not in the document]
Kaszana102: but technically it is correct value, but why would we rotate 20 times, when can none.  It simplyfies multiple rotations.

pkg/tiles/tile.go
Kaszana102: what do you want me to do?

pkg/tiles/feature.go
```diff
@@ -0,0 +1,17 @@
+package tiles
+
+import connection "github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tiles/connection"
+
+type FeatureType int64
+
+const (
+	NONE FeatureType = iota
```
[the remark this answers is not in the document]
Kaszana102: changed

pkg/tiles/feature.go
```diff
@@ -0,0 +1,17 @@
+package tiles
```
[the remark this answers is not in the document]
Kaszana102: moved

pkg/tiles/tile.go
bstrzelecki: > In this case we need to return tile

pkg/tiles/connection/connection.go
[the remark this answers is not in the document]
bstrzelecki: This is just minor suggestion, but there is no actual reason to call this function with rotations > 4 and would be just mistake on callers' part

pkg/tiles/tile.go
Kaszana102: so return the same tile, which was passed as the argument??????

bstrzelecki: Return exact copy of the input tile while only changing tile's connections.
So instead of manually setting every field rely on build in golang value receiver feature.
From https://go.dev/tour/methods/4:
> With a value receiver, the Scale method operates on a copy of the original Vertex value.

Kaszana102: won't it modyfy the original tile?

pkg/tiles/feature/feature.go
```diff
@@ -2,7 +2,7 @@ package feature
 
 import connection "github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tiles/connection"
 
-type FeatureType int64
+type FeatureType int64 //nolint:revive // name type would be too ambigous
```
bstrzelecki, line 5: Currently this identifier will look like this feature.FeatureType, so there is no need to stutter and we can simplify it to feature.Type

Kaszana102: changed

pkg/tiles/tile.go
bstrzelecki: It won't since tile is not a pointer receiver but a value receiver. I would highly encourage you to take a look at the link i posted in this conversation.

Kaszana102: check now

bstrzelecki: Looks ok

bstrzelecki (approved):

pkg/tiles/tile.go
```diff
@@ -0,0 +1,85 @@
...
+/*
+Immutable object
+*/
+type Tile struct {
+	Features  []featureMod.Feature
+	HasShield bool
+	Building  buildings.Bulding
+}
+
+func (tile *Tile) Cities() [][]connectionMod.Side {
```
Jackenmen, line 18: Hmm, wouldn't it be simpler to just return `[]Feature`? Same question for `Roads()` and `Fields()` as well.

pkg/tiles/tile.go
```diff
@@ -0,0 +1,85 @@
...
+	var cities [][]connectionMod.Side
+	for _, feature := range tile.Features {
+		if feature.FeatureType == featureMod.City {
+			cities = append(cities, feature.Connections)
+		}
+	}
+	return cities
+}
+
+func (tile *Tile) CitiesAppendConnection(connections []connectionMod.Side) {
```
Jackenmen, line 28: If the API was accepting `Feature`, we could merge `*AppendConnection()` methods into single `AppendFeature()`.

Jackenmen (changes requested):
Food for thought: `side`/`Sides` and `connection`/`Connections` variable/field names seem to be used interchangeably yet we have both of them in the code, is there a reason behind this distinction?

pkg/tiles/connection/connection.go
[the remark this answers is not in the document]
Jackenmen: While I don't know how this is going to be used but personally, I don't mind a more forgiving API that the current implementation is. It just abstracts away some of the lower-level details for its user.

pkg/tiles/tile.go
Kaszana102: changed

pkg/tiles/tile.go
Kaszana102: as you wish, removed

Jackenmen: I guess you went for having the user use `append(tile.Features, ...)` rather than having a separate method, seems fine.

Jackenmen (approved):
LGTM.

bstrzelecki (approved):

PR #10, opened 2024-04-04
Implemented logic for stack of remaining tiles

### LINKED ISSUE #6: Implement logic for the stack of remaining tiles
At the beginning of the game, the stack of remaining tiles needs to be filled with all tiles (excl. base starting tile) and shuffled (in a well-defined order). Whenever a new turn starts, the player will need to be able to draw the tile from the stack, potentially multiple times until a tile that can be played is drawn.

- Implement a stack structure that can be:
    - popped
    - can be filled with initial data (i.e. all tiles at the beginning of the game)
- Make sure that there is a way to generate the same stack based either on a seed or by creating stack from initial data

bstrzelecki:
closes #6


pkg/stack/tile_stack.go
```diff
@@ -0,0 +1,66 @@
...
+		tiles:   tiles,
+		order:   make([]int32, len(tiles)),
+	}
+	for i := range len(tiles) {
+		stack.order[i] = int32(i)
+	}
+	return stack
+}
+
+func (s *Stack) Shuffle() {
```
Kaszana102, line 37: Why do we need another array of indexes defining order of tiles? Isn't slice alraedy ordered?

pkg/stack/tile_stack.go
```diff
@@ -0,0 +1,66 @@
...
+type Tile struct { // To be replaced by #9
+	id int
+}
+
+type Stack struct {
+	seed    int64
+	turn_no int32
+	tiles   []Tile
+	order   []int32
+}
```
Kaszana102, line 18: It won't be able to function as a discard pile? I mean it doesn't have any function for inserting new tile. Or am i overcomplicating?

Kaszana102 (dismissed):

pkg/stack/tile_stack_test.go
```diff
@@ -0,0 +1,84 @@
...
+	for range 2 {
+		stack.Next()
+	}
+	remainging := stack.GetRemaining()
+	if remainging[0] != tiles[2] {
+		t.Fail()
+	}
+	if remainging[1] != tiles[3] {
+		t.Fail()
+	}
```
Jackenmen, line 83: typo
```suggestion
	remaining := stack.GetRemaining()
	if remaining[0] != tiles[2] {
		t.Fail()
	}
	if remaining[1] != tiles[3] {
		t.Fail()
	}
```

Jackenmen (changes requested):

Kaszana102 (approved):
Ok

pkg/stack/tile_stack.go
```diff
@@ -0,0 +1,77 @@
...
+	turn_no int32
+	tiles   []T
+	order   []int32
+}
+
+type StackOutOfBoundsError struct{}
+
+func (e *StackOutOfBoundsError) Error() string {
+	return fmt.Sprint("stack: out of bounds")
+}
```
Jackenmen, line 20: nit: this could be simplified to:
```suggestion
var StackOutOfBoundsError = errors.New("stack: out of bounds")
```
This is my bad on the terminology since I explicitly asked for a separate error type when I just wanted something that can be easily checked against with i.e. `errors.Is()`

Jackenmen (approved):
LGTM, just one nit that I'm indifferent to.

Jackenmen (approved):
LGTM.

PR #13, opened 2024-04-07
Switch naming convention used in the stack package

Jackenmen:
I missed this during the initial review (#10) but the convention in Go is to use MixedCaps / mixedCaps, not pascal_case, and is also already followed by the rest of the PRs:
https://go.dev/doc/effective_go#mixed-caps


bstrzelecki:
We could consider setting this https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#var-naming as a required check

bstrzelecki (approved):

PR #14, opened 2024-04-07
Add game state abstraction

### LINKED ISSUE #8: Implement abstraction operating on game state
Implement a type responsible for performing operations on the game state (i.e. placing a tile).

Jackenmen:
A lot of method bodies are left as not implemented since it wasn't part of the task. Abstractions should be defined reasonably but the relevant features are yet to be implemented so some things may need to deviate. I covered the code with real test cases that happen to match fake implementation's output. 

Resolves #8


pkg/game/board.go
```diff
@@ -0,0 +1,100 @@
...
+func (board *board) PlaceTile(tile elements.PlacedTile) (elements.ScoreReport, error) {
+	if len(board.tiles) == cap(board.tiles) {
+		return elements.ScoreReport{}, errors.New(
+			"Board's tiles capacity exceeded, logic error?",
+		)
+	}
+	// TODO for future tasks:
+	// - determine if the tile can placed at a given position,
+	//   or return InvalidMove otherwise
+	board.tiles = append(board.tiles, tile)
```
bstrzelecki, line 81: To gain any benefit from having separate structures for holding placed tile data, board.tiles should have deterministic order.

pkg/game/elements/placed_tile.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func (pos *Position) UnmarshalText(text []byte) error {
+	_, err := fmt.Sscanf(string(text), "%v,%v", &pos.x, &pos.y)
+	return err
+}
+
+type Meeple struct {
+	Player Player
+	Side   Side
```
bstrzelecki, line 34: missing meeple type for future proofing

pkg/game/board.go
Jackenmen: The order is deterministic if you know the deck's tile set and seed, guess that's not what you're looking for :smile: 

What kind of order are you looking for? There are multiple identical tiles in the set so I can't simply do a 1:1 mapping, I suppose I could place identical tiles with an increasing offset when getting duplicates, i.e.
```
A, B, and C are different tile configurations, 0 means that the certain position is not filled (only relevant for `board.tiles`)

Tile set: A A A B B B C C

Round 1:
- Placed tile: B
- board.tiles: 0 0 0 B 0 0 0 0

Round 2:
- Placed tile: B
- board.tiles: 0 0 0 B B 0 0 0

Round 3:
- Placed tile: C
- board.tiles: 0 0 0 B B 0 C 0

etc.
```
but I don't know if that's what you're expecting.

bstrzelecki (dismissed):

pkg/game/board.go
```diff
@@ -0,0 +1,126 @@
...
+	// in a slice of size equal to the number of tiles in the set.
+	// `tiles[0]` is always the starting tile.
+	// Indexes in `tiles[1:]` correspond to positions of equivalent tiles in `tileSet`.
+	tiles []elements.PlacedTile
+	// tilesMap is used by the engine for faster lookups
+	// but contains the same information as the `tiles` slice.
+	tilesMap map[elements.Position]elements.PlacedTile
+}
+
+func NewBoard(tileSet []tiles.Tile) elements.Board {
```
BognaLew, line 30: Consider adding starting tile as a parameter. It will be usefull in the future, when we implement River expansion (if a starting tile is a source then we add river tiles).

pkg/game/game.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func (game *Game) PlayerCount() int {
+	return len(game.players)
+}
+
+func (game *Game) ensureCurrentTileHasValidPlacement() error {
+	for {
+		// Peek at the tile that will be returned by GetCurrentTile() next time
+		// to see, if it can actually placed anywhere.
```
BognaLew, line 72:
```suggestion
		// to see, if it can actually be placed anywhere.
```

pkg/game/game.go
```diff
@@ -0,0 +1,172 @@
...
+		nextTile, err := game.deck.Peek()
+		if err != nil {
+			if errors.Is(err, stack.ErrStackOutOfBounds) {
+				// out of bounds - not our concern at the end of a turn
+				return nil
+			}
+
+			// unexpected error?
```
BognaLew, line 80: Maybe Panic?

pkg/game/player.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+type player struct {
+	id uint8
+	// indexed by meeple's enum value
+	meepleCounts []uint8
+	score        uint32
+}
+
+func NewPlayer(id uint8) elements.Player {
```
BognaLew, line 15:
```suggestion
func New(id uint8) elements.Player {
```

pkg/game/player.go
```diff
@@ -0,0 +1,60 @@
+package game
```
BognaLew, line 1: You could move it to separate package.

pkg/logger/log_entries.go
```diff
@@ -1,32 +1,37 @@
...
-	Deck    []int    `json:"deck"`
-	Players []string `json:"players"`
+	Event       string       `json:"event"`
+	Deck        []tiles.Tile `json:"deck"`
+	PlayerCount int          `json:"playerCount"`
 }
 
-func NewStartEntry(deck []int, players []string) StartEntry {
-	return StartEntry{"start", deck, players}
+// TODO: replace with real tile implementation
```
BognaLew, line 15: Tile implementation already exists. Why is it here?

pkg/game/game.go
Jackenmen: Will we be catching panics in game engine or could this crash the whole process during costly machine learning when we could potentially recover (by ignoring that one game during learning process and logging the isssue)?

pkg/game/elements/placed_tile.go
```diff
@@ -0,0 +1,72 @@
...
+	Side   side.Side
+	Type   MeepleType
+}
+
+type LegalMove struct {
+	tiles.Tile
+	Pos Position
+}
+
+type PlacedTile struct {
```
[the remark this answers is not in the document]
Jackenmen: PlacedTile has a Meeple, while LegalMove doesn't which is what I thought was going to the network. If not then we should probably have another method for generating valid moves for a specific player as well.

pkg/game/player.go
Jackenmen: I think this would be overdoing it a bit - `Player` is very closely related with/dependent on the `Game` and `Board` classes. I can ([and did](https://github.com/Jackenmen/Carcassonne-Engine/commit/b20a01aa5b91fb1132ae6b45cef28fa5a2fb3d67)) split it out but I don't think it's actually better than the alternative (and it did mean having to switch `player` tests from tests that can access the package's internals to one that can't - white box -> black box testing, I guess).

https://dave.cheney.net/practical-go/presentations/qcon-china.html#_consider_fewer_larger_packages

pkg/game/game.go
BognaLew: Since we do not know the error it may as well crash the whole process.

pkg/game/elements/placed_tile.go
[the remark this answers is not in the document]
Jackenmen: `LegalMove`s (coordinates + rotation) to put a tile on are player-agnostic but `PlacedTile`s (coordinates + rotation + whether meeple is placed + what meeple is placed) aren't since it depends on whether the player has a certain meeple.

If a list of valid `PlacedTile`s is the input to neural network then we need to have a way to generate these, no? It would be based on existing methods for generating legal moves but would have to instead return a bunch of `PlacedTile`s for each move to include information about all valid meeple placements as well.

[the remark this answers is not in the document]
Jackenmen: Well, it would be more accurate to say that the moves for the player would be a subset of all legal placements of the tile on the board *but* the current `LegalMove` API doesn't make that a possibility either since it doesn't actually contain a `Meeple`...

Yeah, this needs a small redesign...

pkg/deck/deck.go
```diff
@@ -0,0 +1,12 @@
+package deck
+
+import (
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/stack"
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tiles"
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tilesets"
+)
+
+type Deck struct {
```
bstrzelecki, line 9: Isn't this struct a bit redundant? Since both Stack and TIleSet contains complete list of ordered tile

Jackenmen: It would be enough for this struct to have a `StartingTile` and `Stack`, yes. I felt like having the ability to get the whole `TileSet` from this wrapping structure would be useful. You are right that it is a bit redundant though, I'll switch it to a method that constructs `TileSet` from `StartingTile` and `Stack.GetTiles()`.

Jackenmen: Done.

pkg/game/elements/placed_tile.go
```diff
@@ -0,0 +1,112 @@
...
+type LegalMove struct {
+	TilePlacement
+	// LegalMove always has a `Meeple`. Whether it is actually placed
+	// is determined by `MeeplePlacement.Side` which will be `None`, if it isn't.
+	Meeple MeeplePlacement
+}
+
+// note: this is implemented to avoid inheriting the implementation
+// but is just a simple passthrough and its output may not be a legal move.
+func (move LegalMove) Rotate(rotations uint) LegalMove {
```
bstrzelecki, line 87: If calling this method may cause "LegalMove" to no longer be legal then we shouldn't allow this api call to be possible in the first place and just panic out to  break inheritance line.

pkg/game/board.go
```diff
@@ -0,0 +1,134 @@
...
+
+func (board *board) TileCount() int {
+	return len(board.tilesMap)
+}
+
+func (board *board) Tiles() []elements.PlacedTile {
+	return board.tiles
+}
+
+func (board *board) GetTileAt(pos elements.Position) (elements.PlacedTile, bool) {
```
bstrzelecki, line 52:
```suggestion
func (board *board) GetTileAtPosition(pos elements.Position) (elements.PlacedTile, bool) {
```
Just a suggestion

pkg/game/elements/placed_tile.go
Jackenmen: Done.

bstrzelecki (approved):

pkg/game/board.go
Jackenmen: I'd say that it's clear from the signature what `At` is referring to (i.e. calls are likely to be either `board.GetTileAt(pos)` or `board.GetTileAt(elements.NewPosition(...))` where it's clear that we're trying to get a tile at a specified position) and so there's no much point in duplicating that information in the method name.

pkg/game/board.go
```diff
@@ -0,0 +1,134 @@
...
+	// in a slice of size equal to the number of tiles in the set.
+	// `tiles[0]` is always the starting tile.
+	// Indexes in `tiles[1:]` correspond to positions of equivalent tiles in `tileSet`.
+	tiles []elements.PlacedTile
+	// tilesMap is used by the engine for faster lookups
+	// but contains the same information as the `tiles` slice.
+	tilesMap map[elements.Position]elements.PlacedTile
+}
+
+func NewBoard(tileSet tilesets.TileSet) elements.Board {
```
BognaLew, line 31:
```suggestion
func New(tileSet tilesets.TileSet) elements.Board {
```

pkg/game/elements/placed_tile.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tilesets"
+)
+
+type Position struct {
+	// int8 would be fine for base game (72 tiles) but let's be a bit more generous
+	x int16
+	y int16
+}
+
+func NewPosition(x int16, y int16) Position {
```
BognaLew, line 17:
```suggestion
func New(x int16, y int16) Position {
```

BognaLew (approved):

PR #24, opened 2024-04-14
 Implemented logic for processing player's move

### LINKED ISSUE #16: Implement logic for processing player's move
Generate legal moves (+ a way to detect if a move is legal), update the board after player's move, calls to yet-to-be-implemented function for detecting closed features and scoring them.

bstrzelecki:
Closes #16

### DIFF
--- a/pkg/game/board.go
+++ b/pkg/game/board.go
@@ -6,6 +6,8 @@ import (
 
 	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/game/elements"
 	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tiles"
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tiles/feature"
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tiles/side"
 	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tilesets"
 )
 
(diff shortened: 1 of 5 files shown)


BognaLew (approved):

pkg/tiles/tile.go
```diff
@@ -88,3 +90,24 @@ func (tile Tile) Rotate(rotations uint) Tile {
 	tile.Features = newFeatures
 	return tile
 }
+
+func (tile Tile) GetTileRotations() []Tile {
```
Lenetis, line 94: I don't understand what this function does. I'd assume it returns a slice containing all possible rotations of the tile, but why is there a triple nested for loop with two different break statements inside? Is it to eliminate the rotations that would result in the exact same tile? I think a comment explaining exactly what this function returns would be helpful

pkg/game/board.go
```diff
@@ -54,33 +64,86 @@ func (board *board) GetTileAt(pos elements.Position) (elements.PlacedTile, bool)
...
+			tilePlacement := elements.TilePlacement{Tile: currentTile, Pos: placeable}
+			if board.isPositionValid(tilePlacement) {
+				valid = append(valid, tilePlacement)
+			}
+		}
+	}
+	return valid
+}
+
+func (board *board) testSide(position elements.Position, expectedSide side.Side, expectedFeatureType feature.Type) bool {
```
Lenetis, line 81: Please add a comment explaining what this function does and what it returns.

Lenetis (approved):

BognaLew (approved):

PR #25, opened 2024-04-14
Update ScoreReport's player ID keys

Jackenmen:


pkg/game/elements/score_report.go
```diff
@@ -1,6 +1,9 @@
 package elements
 
 type ScoreReport struct {
-	ReceivedPoints  map[int]uint32
-	ReturnedMeeples map[int][]uint8
+	// ReceivedPoints[playerID (uint8) = player's received points
```
Lenetis, line 4:
```
// ReceivedPoints[playerID (uint8) 
Missing a closing ']' ------------^
```

Lenetis (approved):
Everything looks ok, but why are there two getTestScoreReport() functions differing only by first letter case?

pkg/game/elements/score_report.go
Jackenmen: Done.

Jackenmen:
> but why are there two getTestScoreReport() functions differing only by first letter case?

I forgot to remove it when I moved it into a common test utilities package. It's removed now.

Lenetis (approved):

BognaLew (approved):

PR #34, opened 2024-04-20
Replaced plain enum with bitmasks in Side

Lenetis:
Replaced Side values to be a bitmask instead of plain enum. Also removed corners from Side, as they were not going to be used and Center Side, as it is not strictly necessary.
The side values currently are as follows:
```
TopLeftEdge     0b10000000
TopRightEdge    0b01000000

RightTopEdge    0b00100000
RightBottomEdge 0b00010000

BottomRightEdge 0b00001000
BottomLeftEdge  0b00000100

LeftBottomEdge  0b00000010
LeftTopEdge     0b00000001

Top             0b11000000
Right           0b00110000
Bottom          0b00001100
Left            0b00000011

None            0b00000000
```
Rotating the tile has been greatly simplified to use just circular bitshift (aka bitwise rotation) by {2*rotation} bits to the right. (Only the internals of the Rotate function have changed)

Features with multiple sides now use for example `Top | Right` instead of a slice.
Also `Top` is now equal to `TopLeftEdge | TopRightEdge`, etc.

Benefits of this change:
- all features always have the same size (no variable-sized slices)
- features can be compared directly, without the .Equals() method
- faster rotation and comparison (probably, I didn't actually check this :P)
- features can be now used as keys in maps - this is the most important one

### DIFF
--- a/pkg/tiles/side/side.go
+++ b/pkg/tiles/side/side.go
@@ -83,7 +83,7 @@ func (side Side) String() string { //nolint:gocyclo // splitting into multiple s
 /*
 Rotates side clockwise
 */
-func (side Side) Rotate(rotations uint) Side { //nolint:gocyclo // splitting into multiple switches would be obscure
+func (side Side) Rotate(rotations uint) Side {
 	/*
 		Top             0b0000000000000001
 		Right           0b0000000000000010
@@ -115,7 +115,7 @@ func (side Side) Rotate(rotations uint) Side { //nolint:gocyclo // splitting int
 	result := side & ^Center
 
 	for rotations > 0 {
-		result = result << 1
+		result <<= 1
 
 		// loop the last 4 bits (rotated Left goes back into Top)
 		if result&0b1_0000 != 0 {
--- a/pkg/tiles/side/side.go
+++ b/pkg/tiles/side/side.go
@@ -40,7 +40,7 @@ const (
 	None Side = 0
 )
 
-func (side Side) String() string { //nolint:gocyclo // splitting into multiple switches would be obscure
+func (side Side) String() string {
 	sideNames := map[Side]string{
 		Top:             "TOP",
 		Right:           "RIGHT",
--- a/pkg/tiles/side/side.go
+++ b/pkg/tiles/side/side.go
@@ -1,43 +1,38 @@
 package side
 
-type Side int16
+type Side uint8
 
-// The order of sides in this enum matters for some bitwise operations in Rotate().
 const (
-	Top Side = 1 << iota
-	Right
-	Bottom
-	Left
-
-	// for fields
-
 	/* Left side of top edge */
-	TopLeftEdge
-
-	/* Top side of right edge */
-	RightTopEdge
-
-	/* Right side of bottom edge */
-	BottomRightEdge
-
-	/* Bottom side of left edge */
-	LeftBottomEdge
-
-	/* Top side of left edge */
-	LeftTopEdge
+	TopLeftEdge Side = 0b1000_0000
 
 	/* Right side of top edge */
-	TopRightEdge
+	TopRightEdge Side = 0b0100_0000
+
+	/* Top side of right edge */
+	RightTopEdge Side = 0b0010_0000
 
 	/* Bottom side of right edge */
-	RightBottomEdge
+	RightBottomEdge Side = 0b0001_0000
+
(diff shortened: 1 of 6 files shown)


Kaszana102 (approved):
seems ok

bstrzelecki:
Overall change looks great, but this can be simplified, so that  TopLeftEdge | TopRightEdge == Top. Especially since combining this two masks yields invalid state anyway. Also I think we can drop Center side, because by looking at existing tiles this information looks redundant and this change would allow us to neatly fit this in 8-bit integer.

pkg/tiles/side/side.go
```diff
@@ -1,164 +1,144 @@
...
 	BottomLeftEdge
-	/* Right side of bottom edge */
-	BottomRightEdge
+
+	Center
+
+	None Side = 0
 )
 
 func (side Side) String() string { //nolint:gocyclo // splitting into multiple switches would be obscure
```
bstrzelecki, line 43: Linter directive should no longer be applicable

Kaszana102 (dismissed):

Lenetis:
> Overall change looks great, but this can be simplified, so that TopLeftEdge | TopRightEdge == Top. Especially since combining this two masks yields invalid state anyway. Also I think we can drop Center side, because by looking at existing tiles this information looks redundant and this change would allow us to neatly fit this in 8-bit integer.

Both of these are used in our MonasteryWithSingleRoad() tile template, and TopLeftEdge | TopRightEdge isn't equal to Top there:

image

Also Center side would be necessary if we want to implement this tile:

image

bstrzelecki:
> Both of these are used in our MonasteryWithSingleRoad() tile template, and TopLeftEdge | TopRightEdge isn't equal to Top there:
> 
> image

In this example:
- We can infer that road with single connection is terminated
- Monastery does not rely on being in specific side so this can be disregarded
- Road has sides: Top or TopLeftEdge and TopRightEdge
- Field has sides: Top, Bottom, Left and Right or all respective edges


> Also Center side would be necessary if we want to implement this tile:
> 
> image

This can be modeled as Field having "None" external connections

Lenetis:
> > Both of these are used in our MonasteryWithSingleRoad() tile template, and TopLeftEdge | TopRightEdge isn't equal to Top there:
> > image
> 
> In this example:
> 
>     * We can infer that road with single connection is terminated
> 
>     * Monastery does not rely on being in specific side so this can be disregarded
> 
>     * Road has sides: Top or TopLeftEdge and TopRightEdge
> 
>     * Field has sides: Top, Bottom, Left and Right or all respective edges
> 
> 
> > Also Center side would be necessary if we want to implement this tile:
> > image
> 
> This can be modeled as Field having "None" external connections

Generally I agree and it's a good idea, but what about meeple placement? If we model the second example as field having a None side, how can we place a meeple there?
```
// LegalMove always has a `Meeple`. Whether it is actually placed
// is determined by `MeeplePlacement.Side` which will be `None`, if it isn't.
```
My ideas for solving this:
- add bool HasMeeple to LegalMove
- stick to uint16 and add the Center side as a one extra bit
- treat 0b11111111 as a special case in MeeplePlacement and use that as center (weird solution but the most efficient probably, since a meeple cannot be placed on all sides at once)

bstrzelecki:
> Generally I agree and it's a good idea, but what about meeple placement? If we model the second example as field having a None side, how can we place a meeple there?

Meeple placement is subject to change as per #27, such that meeple position will be linked with feature directly rather than being on specific side

pkg/tiles/side/side.go
```diff
@@ -1,164 +1,142 @@
...
-	TopLeftCorner
-	TopRightCorner
-	BottomLeftCorner
-	BottomRightCorner
-
-	// for fields
-
 	/* Left side of top edge */
-	TopLeftEdge
+	TopLeftEdge Side = 0b1000_0000
```
bstrzelecki, line 7: This still could be defined using iota

bstrzelecki (approved):

pkg/tiles/side/side.go
Lenetis: I know, but I thought this would be more explicit and understandable.
The sides Top, Right, Bottom, Left still need to be written as 0b... so having the above ones in the same format makes it more clear which bits correspond to which side

Kaszana102 (approved):

PR #37, opened 2024-04-28
implented scoring roads

Kaszana102:
fixing rebase from other branch


Lenetis (dismissed):

pkg/game/board.go
```diff
@@ -132,3 +134,167 @@ func (board *board) checkCompleted(
 	}
 	return scoreReport, nil
 }
+
+/*
+Tt doesn't analyze starting tile.
```
bstrzelecki, line 139: Tt?

pkg/game/board.go
```diff
@@ -132,3 +134,167 @@ func (board *board) checkCompleted(
...
+	}
+
+	finished = (road.Sides.GetCardinalDirectionsLength() == 1) || (tile.Pos == startTile.Pos)
+	finished = finished && (err == nil) && tileExists
+
+	looped := (tile.Pos == startTile.Pos) && singleIterationMade
+	return finished, score, meeples, looped
+}
+
+func CreateScoreRoadReport(score int, meeples []elements.MeepleTilePlacement) elements.ScoreReport {
```
bstrzelecki, line 204: Should be located in score_report.go

pkg/game/board.go
```diff
@@ -132,3 +134,167 @@ func (board *board) checkCompleted(
...
+			}
+
+			// check for meeple2 (other end of road)
+			if tile.Meeple.Side == roadSide {
+				meeples = append(meeples, elements.MeepleTilePlacement{MeeplePlacement: tile.Meeple, PlacedTile: tile})
+			}
+		}
+
+		if road.Sides.GetCardinalDirectionsLength() == 1 || err != nil {
+
```
bstrzelecki, line 192:
```suggestion
```

pkg/game/board.go
```diff
@@ -132,3 +134,167 @@ func (board *board) checkCompleted(
...
+	// check finished on way
+	// do while loop
+	for {
+		singleIterationMade = true
+		tile, tileExists = board.GetTileAt(tile.Pos.Add(elements.PositionFromSide(roadSide)))
+		// check if tile exists or loop
+		if !tileExists || tile.Pos == startTile.Pos {
+			// tile does not exist
+			// finish
+
```
bstrzelecki, line 161:
```suggestion
```

pkg/game/board.go
```diff
@@ -132,3 +134,167 @@ func (board *board) checkCompleted(
...
+	if !loopResult && rightSide != side.None {
+		roadFinishedResult, scoreResult, meeplesResult, _ = board.CheckRoadInDirection(rightSide, tile)
+		score += scoreResult
+		roadFinished = roadFinished && roadFinishedResult
+		meeples = append(meeples, meeplesResult...)
+	}
+
+	// -------- start counting -------------
+	if roadFinished {
+
```
bstrzelecki, line 281:
```suggestion
```

pkg/game/elements/placed_tile.go
```diff
@@ -26,6 +26,44 @@ func (pos Position) Y() int16 {
 	return pos.y
 }
 
+func (pos Position) Add(other Position) Position {
+	return NewPosition(pos.x+other.x, pos.y+other.y)
+}
+
+func PositionFromSide(side sideMod.Side) Position { //nolint:gocyclo // similar problem to sides
```
bstrzelecki, line 33: This function should use bitmask operations, as it will also make nolint directive redundant

pkg/game/test/factories.go
```diff
@@ -45,3 +45,16 @@ func GetTestScoreReport() elements.ScoreReport {
 		ReturnedMeeples: map[uint8][]uint8{},
 	}
 }
+
+func GetTestTestPlacedTile(tileTemplate tiles.Tile, playerID uint8) elements.PlacedTile {
```
bstrzelecki, line 49: This function stutters

pkg/tiles/side/side.go
```diff
@@ -140,3 +142,73 @@ func (side Side) Rotate(rotations uint) Side {
 	var shift = rotations * 2
 	return (side >> shift) | (side << (8 - shift)) // circular bitshift (bitwise rotate) side to the right by {2*rotations} bits
 }
+
+/*
+argument indicates only ONE cardinal or edge, otherwise it's ambigous
+*/
+func (side Side) ConnectedOpposite() (Side, error) { //nolint:gocyclo // splitting into multiple switches would be obscure
```
bstrzelecki, line 149: This function should use bitmask operations, as it will also make nolint directive redundant and would let us  avoid returning error

pkg/tiles/side/side.go
```diff
@@ -140,3 +142,73 @@ func (side Side) Rotate(rotations uint) Side {
...
+		return TopLeftEdge, nil
+	case BottomRightEdge:
+		return TopRightEdge, nil
+
+	default:
+		return None, errors.New("side None side has not opposite")
+	}
+}
+
+func RotateSideArray(sides []Side, rotations uint) []Side {
```
bstrzelecki, line 183: Is it still required?

pkg/game/board.go
Kaszana102: It*

pkg/game/board.go
Kaszana102: function moved and renamed

pkg/game/test/factories.go
Kaszana102: renamed

pkg/tiles/side/side.go
Kaszana102: it seems not.

pkg/tiles/side/side.go
Kaszana102: changed

pkg/game/elements/placed_tile.go
Kaszana102: is it ok now?

pkg/game/scoring_roads_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+	// add meeple to first road
+	tiles[0].GetPlacedFeatureAtSide(side.Right).Meeple = elements.Meeple{PlayerID: 1, MeepleType: elements.NormalMeeple}
+
+	// set positions
+	tiles[0].Position = elements.NewPosition(-1, 0)
+	tiles[1].Position = elements.NewPosition(1, 0)
+
+	expectedScores := []uint32{0, 3}
+	expectedMeeples := [][]uint8{nil, {0, 1}} // expected zero Nones, and 1 normal meeple
```
bstrzelecki, line 154: Why do we need to track the amount of "None" meeples in score report?

pkg/game/board.go
```diff
@@ -222,3 +222,140 @@ func (board *board) checkCompleted(
 	}
 	return scoreReport, nil
 }
+
+/*
+It doesn't analyze starting tile.
```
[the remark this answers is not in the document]
Kaszana102: Reordered

pkg/game/board.go
```diff
@@ -222,3 +222,140 @@ func (board *board) checkCompleted(
...
+param roadSide: always indicates only one cardinal direction!
+*/
+func (board *board) CheckRoadInDirection(roadSide side.Side, startTile elements.PlacedTile) (bool, int, []elements.Meeple, bool, side.Side) {
+	var meeples = []elements.Meeple{}
+	var tile = startTile
+	var tileExists bool
+	var score = 0
+	var road *elements.PlacedFeature
+	var finished bool
+	var singleIterationMade = false // to prevent ending before entering loop (f.e.: placed tile is a monastery with a road, so one side is Center from the beginning but it's not loop)
```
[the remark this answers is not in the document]
Kaszana102: thank you for that! I realized that, but forgot to remove it!

pkg/game/scoring_roads_test.go
Kaszana102: Actually i didn't know how to create map without the None meeple key. I create the map in score_report.go in line 90.  Is there better way to achieve this? 
There is something like this:
scoreReport.ReturnedMeeples[uint8(meeple.PlayerID)] = []uint8{0, 0}

When i did it with one zero, it didn't work.

pkg/tiles/side/side.go
```diff
@@ -140,3 +140,53 @@ func (side Side) Rotate(rotations uint) Side {
...
+
+	// mirror edges
+	for i := range 4 {
+		rotated = SwapBits(rotated, uint8(2*i), uint8(2*i+1))
+	}
+	return rotated
+}
+
+func (side Side) GetNthCardinalDirection(n uint8) Side {
+	cardinals := []Side{Top, Left, Right, Bottom}
```
Lenetis, line 170: Please add a comment explaining what this function does.
Also, does it check the directions in a specific order, and does the order matters for anything? Why?

pkg/game/board.go
```diff
@@ -222,3 +222,138 @@ func (board *board) checkCompleted(
...
+		if road.Sides.GetNthCardinalDirection(0) == roadSide {
+			roadSide = road.Sides.GetNthCardinalDirection(1)
+		} else {
+			roadSide = road.Sides.GetNthCardinalDirection(0)
+		}
+
+	}
+	finished = tileExists && ((road.Sides.GetCardinalDirectionsLength() == 1) || (tile.Position == startTile.Position))
+
+	looped := (tile.Position == startTile.Position)
```
Lenetis, line 275: I'd reorder this to:
```
looped := (tile.Position == startTile.Position)

finished = tileExists && ((road.Sides.GetCardinalDirectionsLength() == 1) || looped)
```

pkg/game/board.go
```diff
@@ -222,3 +222,138 @@ func (board *board) checkCompleted(
...
+	// check meeples on start tile
+	var roadLeft = tile.GetPlacedFeatureAtSide(leftSide)
+	var roadRight = tile.GetPlacedFeatureAtSide(rightSide)
+	if roadLeft.MeepleType != elements.NoneMeeple || (roadRight != nil && roadRight.MeepleType != elements.NoneMeeple) {
+		if roadLeft.MeepleType != elements.NoneMeeple {
+			meeples = append(meeples, roadLeft.Meeple)
+		} else {
+			meeples = append(meeples, roadRight.Meeple)
+		}
+	}
```
Lenetis, line 307: Please consider rewriting this to:
```
if roadLeft.MeepleType != elements.NoneMeeple {
	meeples = append(meeples, roadLeft.Meeple)
} else if roadRight != nil && roadRight.MeepleType != elements.NoneMeeple {
	meeples = append(meeples, roadRight.Meeple)
}
```
(shorter, doesn't have a big compound if condition, and doesn't check if roadLeft has a meeple twice)

pkg/game/board.go
```diff
@@ -222,3 +222,138 @@ func (board *board) checkCompleted(
...
 	return scoreReport, nil
 }
+
+/*
+It analyzes road directed by roadSide parameter.
+It doesn't analyze starting tile.
+returns: road_finished, score, [meeples on road], loop, sideFinishedOn
+param roadSide: always indicates only one cardinal direction!
+*/
+func (board *board) CheckRoadInDirection(roadSide side.Side, startTile elements.PlacedTile) (bool, int, []elements.Meeple, bool, side.Side) {
```
Lenetis, line 232: What is "sideFinishedOn"? In ScoreRoadCompletion it's assigned to a variable called "loopSide", but it's not clear to me what it means

pkg/game/elements/score_report.go
```diff
@@ -7,3 +7,90 @@ type ScoreReport struct {
...
+	report.ReceivedPoints = make(map[uint8]uint32)
+	report.ReturnedMeeples = make(map[uint8][]uint8)
+	return report
+}
+
+/*
+Create score report by checking meeples control on the same feature, getting rid of not scoring meeples.
+Returns a score report
+*/
+func CalculateScoreReportOnMeeples(score int, meeples []Meeple) ScoreReport {
```
Lenetis, line 57: "checking meeples control on the same feature" - it's not actually a single feature but a thing comprised of multiple features. Consider changing the name to something more fitting
Also I'd suggest changing "getting rid of" to "ignoring" because we are not actually getting rid of any meeples

pkg/game/board.go
```diff
@@ -222,3 +222,138 @@ func (board *board) checkCompleted(
...
+Calculates score for road.
+
+returns: ScoreReport, checked sides of the start tile (also including loop)
+*/
+func (board *board) ScoreRoadCompletion(tile elements.PlacedTile, road feature.Feature) (elements.ScoreReport, side.Side) {
+	var meeples = []elements.Meeple{}
+	var leftSide, rightSide side.Side
+	var score = 1
+	leftSide = road.Sides.GetNthCardinalDirection(0)
+	rightSide = road.Sides.GetNthCardinalDirection(1)
```
Lenetis, line 289: If I understand correctly, these sides will not always be left and right, but for example top and left or something. I'd suggest adding a comment to explain it (or maybe changing the variable names, but I think comment would be enough)

pkg/game/elements/placed_tile.go
```diff
@@ -93,3 +129,12 @@ type PlacedTile struct {
...
 }
+
+func (placement *PlacedTile) GetPlacedFeatureAtSide(sideToCheck sideMod.Side) *PlacedFeature {
+	for i, feature := range placement.TileWithMeeple.Features {
+		if sideToCheck&feature.Sides == sideToCheck {
+			return &placement.TileWithMeeple.Features[i]
+		}
+	}
+	return nil
+}
```
Lenetis, line 140: Will this behave correctly with tiles that have more than one feature on one side?
For example, monastery with a single road has both a road and a field on one side

pkg/tiles/tile.go
```diff
@@ -83,6 +87,15 @@ func (tile Tile) Rotate(rotations uint) Tile {
...
 }
 
+func (tile *Tile) GetFeatureAtSide(sideToCheck sideMod.Side) *featureMod.Feature {
+	for _, feature := range tile.Features {
+		if sideToCheck&feature.Sides == sideToCheck {
+			return &feature
+		}
+	}
+	return nil
+}
```
Lenetis, line 97: Will this behave correctly with tiles that have more than one feature on one side?
For example, monastery with a single road has both a road and a field on one side

pkg/game/test/factories.go
```diff
@@ -21,3 +21,25 @@ func GetTestScoreReport() elements.ScoreReport {
 		ReturnedMeeples: map[uint8][]uint8{},
 	}
 }
+
+func GetTestCustomPlacedTile(tileTemplate tiles.Tile, playerID uint8) elements.PlacedTile {
+	var placedFeatures []elements.PlacedFeature
+
+	// covnert features to placedFeature
```
Lenetis, line 28:
```suggestion
	// convert features to placedFeature
```

pkg/game/test/factories.go
```diff
@@ -21,3 +21,25 @@ func GetTestScoreReport() elements.ScoreReport {
...
+		})
+	}
+
+	return elements.PlacedTile{
+		TileWithMeeple: elements.TileWithMeeple{
+			Features:  placedFeatures,
+			HasShield: false,
+		},
+		Position: elements.NewPosition(0, 0),
+	}
```
Lenetis, line 44: I think HasShield is no longer required in elements.PlacedTile, because shield is now a feature modifier.
@bstrzelecki did we miss this in #38 ?

pkg/game/scoring_roads_test.go
bstrzelecki: ReturnedMeeples field is defined as map of slices, thus slice with single value should be valid.

pkg/game/test/factories.go
bstrzelecki: Yup, this field is redundant as of #38.

pkg/game/scoring_roads_test.go
Kaszana102: Yea, but NormalMeeple has value 1, and NoneMeeple 0. Thus checking how many certain meeple type did player have, requires substracting by one, 
F.e: 
scoreReport.ReturnedMeeples[uint8(meeple.PlayerID)][meeple.MeepleType-1]++
instead of
scoreReport.ReturnedMeeples[uint8(meeple.PlayerID)][meeple.MeepleType]++

which would be obscure.  I'd rather change it to map of maps, to avoid it, but still in different pull request.

pkg/tiles/side/side.go
```diff
@@ -140,3 +140,53 @@ func (side Side) Rotate(rotations uint) Side {
 	var shift = rotations * 2
 	return (side >> shift) | (side << (8 - shift)) // circular bitshift (bitwise rotate) side to the right by {2*rotations} bits
 }
+
+/*
+argument indicates only ONE cardinal or edge, otherwise it's ambigous
+*/
+func (side Side) ConnectedOpposite() Side {
```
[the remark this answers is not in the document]
Kaszana102: commented

pkg/game/board.go
```diff
@@ -222,3 +222,138 @@ func (board *board) checkCompleted(
...
+	// check finished on way
+	// do while loop
+	for {
+		tile, tileExists = board.GetTileAt(tile.Position.Add(elements.PositionFromSide(roadSide)))
+		roadSide = roadSide.ConnectedOpposite()
+		// check if tile exists or loop
+		if !tileExists || tile.Position == startTile.Position {
+			// tile does not exist
+			// finish
+			break
```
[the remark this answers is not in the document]
Kaszana102: updated comment

pkg/game/board.go
```diff
@@ -222,3 +222,138 @@ func (board *board) checkCompleted(
...
+		if road.Sides.GetCardinalDirectionsLength() == 1 {
+			// found the end of a road
+			break
+		}
+		// swap to other end of the road on the same tile
+		if road.Sides.GetNthCardinalDirection(0) == roadSide {
+			roadSide = road.Sides.GetNthCardinalDirection(1)
+		} else {
+			roadSide = road.Sides.GetNthCardinalDirection(0)
+		}
```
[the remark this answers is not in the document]
Kaszana102: Getting the other side of the road, to indicate where to look for the next tile

pkg/tiles/side/side.go
Kaszana102: commented, checking is in specific order, but reordering it wouldn't change anything. (iterating on all sides simply always in same order)

pkg/game/board.go
Kaszana102: reodered

pkg/game/board.go
Kaszana102: rewritten

pkg/game/board.go
Kaszana102: Added more comments.
It is used for preventing checking the same road twice, as in test TestBoardScoreRoadLoopCrossroad, where the board looks like this:
  - 0 -
    1 - 2
    |	  |
  - 4 - 3
Fourth tile was placed last, and normally the road would be counted twice, but sideFinishedOn is passed to the function ScoreRoads to indicated that this is the same road, from another side.

pkg/game/board.go
Kaszana102: It's much easier to ilustrate with those names, it is supposed to indicate both ends, but naming it first and second end may lead to harder to follow the code.

pkg/game/elements/placed_tile.go
Kaszana102: you're right, the feature type also needs to be passed

pkg/game/test/factories.go
Kaszana102: fixed

pkg/game/test/factories.go
Kaszana102: Am i supposed to do something with it, or what? It's not this pull request for.

pkg/tiles/side/side.go
Lenetis: Ok, I understand, but am still not sure about the order. For example, if `side = Left | Bottom`, side.GetNthCardinalDirection(1) would return Bottom, but with a different iteration order it could return Top.

I think what you really need instead of this function is something like "side.GetConnectedCardinalDirection(direction)" (or GetOtherCardinalDirection, something like that), which in this example would behave like so:
side.GetConnectedCardinalDirection(Left) = Bottom
side.GetConnectedCardinalDirection(Bottom) = Left

This could also allow rewriting this:
```
if road.Sides.GetNthCardinalDirection(0) == roadSide {
	roadSide = road.Sides.GetNthCardinalDirection(1)
} else {
	roadSide = road.Sides.GetNthCardinalDirection(0)
}
```
to a much simpler:
```
roadSide = road.Sides.GetConnectedCardinalDirection(roadSide)
```

pkg/game/test/factories.go
Lenetis: I think if it doesn't cause any problems, we can leave it for now

pkg/game/test/factories.go
```diff
@@ -21,3 +21,25 @@ func GetTestScoreReport() elements.ScoreReport {
...
+
+	// convert features to placedFeature
+	for _, feature := range tileTemplate.Features {
+		placedFeatures = append(placedFeatures, elements.PlacedFeature{
+			Feature: feature,
+			Meeple: elements.Meeple{
+				MeepleType: elements.NoneMeeple,
+				PlayerID:   elements.ID(playerID)},
+		})
+	}
```
Lenetis, line 36: I'd suggest removing playerID from this function and just setting the Meeple's PlayerID to NonePlayer - currently it assigns a player to *every* feature on this tile (albeit meepleless), and later on you set the playerID manually anyways

Kaszana102: removed

pkg/tiles/side/side.go
Kaszana102: I left the function GetNthCardinalDirection, as it simplifies the beggining of the ScoreRoadCompletion function. I created the new function GetConnectedOtherCardinalDirection, which you described.

Lenetis (approved):

pkg/game/scoring_roads_test.go
Lenetis: I'm facing the same problem now and I'd also prefer the non-obscure approach of just having a 0 at the first index to allow easy ignoring of None meeples

PR #38, opened 2024-05-25
Meeple's position identified by the Feature rather than the Side

### LINKED ISSUE #27: Make meeple's position be identified by the Feature rather than the Side

bstrzelecki:
Closes #27


pkg/game/elements/placed_tile.go
```diff
@@ -39,51 +40,49 @@ func (pos *Position) UnmarshalText(text []byte) error {
...
-// represents a legal move (tile placement and meeple placement) on the board
-type LegalMove struct {
-	TilePlacement
-	// LegalMove always has a `Meeple`. Whether it is actually placed
-	// is determined by `MeeplePlacement.Side` which will be `None`, if it isn't.
-	Meeple MeeplePlacement
+func ToPlacedTile(tile tiles.Tile) PlacedTile {
+	features := []PlacedFeature{}
+	for _, n := range tile.Features {
+		features = append(features, PlacedFeature{n, NoneMeeple, 255})
```
Lenetis, line 68: Probably out of scope for this PR, but I think we should change playerID type from uint8 to a dedicated type.
The 255 is kind of a magic number and something like playerID.NonePlayer would be more intuitive (as I assume that's the intent of 255 here)

pkg/game/game.go
```diff
@@ -100,20 +100,10 @@ func (game *Game) ensureCurrentTileHasValidPlacement() error {
...
 	// or `OutOfBounds` error, if there's no tiles left in the deck and this turn
 	// shouldn't be happening.
-	currentTile, err := game.GetCurrentTile()
-	if err != nil {
-		return err
-	}
-
-	if !currentTile.Equals(move.Tile) {
-		return elements.ErrWrongTile
-	}
```
Lenetis, line 115: Why remove these lines?

pkg/player/player.go
```diff
@@ -54,30 +53,40 @@ func (player *player) GetEligibleMovesFrom(moves []elements.LegalMove) []element
...
-	if move.Meeple.Side != side.None {
-		meepleCount := player.MeepleCount(move.Meeple.Type)
-		player.SetMeepleCount(move.Meeple.Type, meepleCount-1)
+	for _, feature := range move.Features {
+		if feature.MeepleType != elements.NoneMeeple {
+			if player.MeepleCount(feature.MeepleType) != 0 {
+				meepleCount := player.MeepleCount(feature.MeepleType)
+				player.SetMeepleCount(feature.MeepleType, meepleCount-1)
+			}
+		}
```
Lenetis, line 89: I'm not sure if we should check here if the player's meeple count is != 0. A situation when it is 0 should never happen.
However, if it may happen, it shouldn't be silently ignored. I'd advise for either removing this if, or adding `else{ panic() }` below.

Lenetis (commented):
Generally ok.
In the future, monasteries should also be changed to features, but I don't insist on doint that in this PR.

Lenetis (approved):

BognaLew (approved):

PR #42, opened 2024-07-09
Moved things related to positions from "placed_tile.go" to "position.go"

Lenetis:


bstrzelecki:
Since position in now in a separate file, we can separate this into its own package, because its most commonly used feature anyway.

BognaLew (changes requested):
Please add tests for Add and PositionFromSide.

pkg/game/elements/position.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (pos Position) Add(other Position) Position {
+	return NewPosition(pos.x+other.x, pos.y+other.y)
+}
+
+/*
+Returns relative position directed by the side.
+Caution! It is supposed to be used with side directing only one cardinal direction (or two edges connected by corner)!
+Otherwise it will return undesired value!
```
bstrzelecki, line 34: We might also want to handle the case when the function would return undesired value.

bstrzelecki (changes requested):

pkg/game/elements/position.go
Lenetis: TBH, I'm not a fan of this idea. It adds a lot of complexity to a very simple function.
Maybe that's just a premature optimisation thought, but I have a feeling iterating through all primary sides to check if more than one is set everytime we call this function would make the engine run slower.

Lenetis:
- Moved position to its own package
- Added tests
- Simplified the position.FromSide() function - I used an older version of this function and I'm not sure why it was changed. Please take a look at commit [3969cb3](https://github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pull/42/commits/3969cb396f4b80af437acf86716180dcedb86ea5) because maybe I'm overlooking some edge case here.

pkg/game/elements/position.go
bstrzelecki: Well, I'm not a fan of methods that require more lines of documentation designated for how this function can fail than how it's actually used. This method incorrectly handles invalid cases and minimal solution would be to just test for strict equallity (i.e. side == sidemod.Top) so that the default case would be actually useful where one could panic or return identity. Test cases on how invalid cases are handled would also be welcomed.

bstrzelecki: Well, I'm not a fan of methods that require more lines of documentation designated for how this function can fail than how it's actually used. This method incorrectly handles invalid cases and minimal solution would be to just test for strict equallity (i.e. side == sidemod.Top) so that the default case would be actually useful where one could panic or return identity. Test cases on how invalid cases are handled would also be welcomed.

Lenetis: Added panic on incorrect input, and updated the documentation comment to better reflect what the function actually does

pkg/game/position/position.go
```diff
@@ -0,0 +1,70 @@
...
+		case side&sideMod.Right != 0:
+			return New(1, 0)
+		case side&sideMod.Left != 0:
+			return New(-1, 0)
+		case side&sideMod.Bottom != 0:
+			return New(0, -1)
+		}
+	}
+	panic(fmt.Sprintf("position.FromSide called with more than one primary side. 'side' = %08b", side))
+}
```
bstrzelecki, line 61:
```suggestion
func FromSide(side sideMod.Side) Position {
	switch side {
	case sideMod.Top:
		return New(0, 1)
	case sideMod.Right:
		return New(1, 0)
	case sideMod.Left:
		return New(-1, 0)
	case sideMod.Bottom:
		return New(0, -1)
	default:
		panic(fmt.Sprintf("position.FromSide called with more than one primary side. 'side' = %08b", side))
	}
}
```

Lenetis: This won't work as intended, as for example (0, 1) should be returned not only when side=Top, but also when it's equal to TopRightCorner and TopLeftCorner. That's why there is & and checking if it's != 0

(see the documentation comment for more details)

bstrzelecki: Ok, then only tests that feature invalid cases (e.g. combination of sideMod.Left and sideMod.BottomRightEdge) are missing.

Lenetis: They are not missing.
In the `TestPositionFromSidePanic` function (in position_test.go) the following invalid cases are tested:
```
side.Top | side.Left,
side.RightBottomEdge | side.BottomRightEdge,
side.Top | side.Right | side.Bottom | side.Left,
side.Top | side.BottomLeftEdge,
```

bstrzelecki (approved):

PR #47, opened 2024-07-15
Logger update

BognaLew:


bstrzelecki (approved):

pkg/logger/logger.go
```diff
@@ -13,13 +13,19 @@ func New(writer io.Writer) Logger {
 	return Logger{writer}
 }
 
-func (logger *Logger) LogEvent(event interface{}) error {
+func (logger *Logger) LogEvent(eventName string, event interface{}) error {
```
Kaszana102, line 16: Honestly i would create some enum with toString method, for limiting possible names. It will make a list of all possible event names instead of searching by hand. (Replace eventName string to eventName someEnum)

Kaszana102 (changes requested):

Kaszana102 (approved):

bstrzelecki (approved):

PR #55, opened 2024-07-25
Removing meeples

Kaszana102:


BognaLew (commented):
Does it solve the issue with end to end tests? If you haven't checked that then please do a rebase the branch with tests on top of this branch and see if the tests pass.

Lenetis (commented):
This is way more complicated than it needs to be. There can be no more than one meeple per tile, so we don't need to track the meeple's type, side and feature in order to remove it from the board.

If it works, it works, but personally I would just add a "MeeplePositions" field to ScoreReport (just a []Position slice), without drastically changing any of the feature scoring code.

Kaszana102:
> This is way more complicated than it needs to be. There can be no more than one meeple per tile, so we don't need to track the meeple's type, side and feature in order to remove it from the board.
> 
> If it works, it works, but personally I would just add a "MeeplePositions" field to ScoreReport (just a []Position slice), without drastically changing any of the feature scoring code.

Hmm, that's a rather good idea.

Kaszana102:
Simplified as krzysztof said, and removing was tested on end tests branch. Works properly. It depends only on correct reports

Lenetis (approved):
IMO the type `MeepleWithPosition` is unnecessary as we need only the meeples' positions, and don't need linking the positions to the meeples 'type. A simple `meeplePositions []Position` slice in ScoreReport could solve this, without the need to rewrite all scoring functions.
But if it works, it works

BognaLew (approved):

PR #58, opened 2024-07-31
Binary tiles

### LINKED ISSUE #39: Implement a binary Tile representation
Currently Tiles are represented as a variable-length list of Features. While this works well for the game engine itself, it will most likely cause problems when trying to feed this data into a neural network. Therefore, a binary tile representation would be useful.

-----

My idea for solving this problem:
- Every main Feature (city, field, road (+ river in expansions)) is treated as a 3x3 grid + extra bits
- Every Tile is comprised of one such grid per feature type (so 3 grids without expansions) and a few extra bits

-----

Feature grids:
- 4 "feature" bits for each side (roads and cities), or for each corner (fields), corresponding to whether a feature includes the given side/corner
- 6 "connection" bits, representing connections between the "feature" bits - needed to determine which sides/corners are connected within the same feature, and which belong to distinct features

image

* F - Feature bit
* C - Connection bit

Examples:

image

image

image

image

-----

Additional bits:
- Monastery represented by a single bit
- Field that is not connected to any corner (for example when it is surrounded by cities from all sides) represented as an additional bit
- Shield and other modifiers represented by 4 bits, one for each side. If a feature with a modifier occupies more than one side, the modifier bit is set on every side of this feature
- Meeples represented as 9 bits - one bit for each side and for each corner + additional "center" bit for a monastery or unconnected field. If there is a meeple on a feature occupying more than one side/corner, the meeple bit is set on every side/corner it occupies

Final example:

image

-----

In summary:
- In a game without expansions, each tile has:
	- 3 main features (field, road, city) = 30 bits
	- 1 monastery bit and 1 unconnected field bit = 2 bits
	- 4 shield bits
	- 9 meeple bits
- In total, each tile is 45 bits long. I assume we need to pad this to 48 bits (6 bytes), so we have 3 additional, unused bits. One of them could be used as a short "is placed" flag - 0 in all empty (not placed) tiles and 1 in non-empty ones. This could probably improve the neural network performance or learning speed. The other two (or alternatively all three) could be used to represent the tile "owner" - 0 in tiles without a Meeple and a number corresponding to a player (2-bit player ID or a one-hot encoding of a player (though one-hot limits max number of players to two or three)) in tiles where a Meeple is present.

-----

Expansions and scaling:
- Each additional main feature (such as river): +10 bits
- Each additional modifier (assuming modifiers are only in cities): +4 bits
- Each additional Meeple type: +1 bit
- Increasing max players number (assuming one-hot encoding): +1 bit (per player)
- Other changes: ???

Lenetis:
Resolves #39

### DIFF
--- a/pkg/game/elements/placed_tile.go
+++ b/pkg/game/elements/placed_tile.go
@@ -25,7 +25,7 @@ type Meeple struct {
 
 type TileWithMeeple struct {
 	Features  []PlacedFeature
-	HasShield bool
+	HasShield bool // todo should be removed
 }
 
 type PlacedFeature struct {
--- a/pkg/game/test/factories.go
+++ b/pkg/game/test/factories.go
@@ -23,7 +23,7 @@ func GetTestScoreReport() elements.ScoreReport {
 	}
 }
 
-func GetTestCustomPlacedTile(tileTemplate tiles.Tile) elements.PlacedTile {
+func GetTestCustomPlacedTile(tileTemplate tiles.Tile) elements.PlacedTile { // todo do we need this function? There is already elements.ToPlacedTile
 	var placedFeatures []elements.PlacedFeature
 
 	// convert features to placedFeature
--- a/pkg/tiles/binarytiles/binary_tile.go
+++ b/pkg/tiles/binarytiles/binary_tile.go
@@ -67,7 +67,6 @@ var connectionMasks = []int64{
 
 func fromPlacedFeatures(features []elements.PlacedFeature) BinaryTile {
 	var binaryTile BinaryTile
-	var owner elements.ID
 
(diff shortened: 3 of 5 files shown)


Kaszana102 (approved):
I couldn't find a tile which couldn't be described this way.

pkg/game/elements/placed_tile.go
```diff
@@ -25,7 +25,7 @@ type Meeple struct {
 
 type TileWithMeeple struct {
 	Features  []PlacedFeature
-	HasShield bool
+	HasShield bool // todo should be removed
```
bstrzelecki, line 28: I reckon we can actually remove those redundant elements.

PR #61, opened 2024-08-03
Implement game tree requests

### LINKED ISSUE #60: Implement and expose game tree generation in game engine
Input: state of the game
Output: all possible moves (based on remaining tiles in the deck) from

Jackenmen:
Fixes #60


python_bindings/tests/engine_test.py
```diff
@@ -85,3 +91,102 @@ def test_game_engine_send_batch_returns_failures_when_communicator_closed(
...
+        StartingTile=tiletemplates.SingleCityEdgeStraightRoads(),
+        Tiles=Slice_tiles_Tile(tiles),
+    )
+
+    engine = StartGameEngine(1, str(tmp_path))
+    game = engine.GenerateGame(tile_set)
+
+    request = GetRemainingTilesRequest(BaseGameID=game.ID)
+    resp, = engine.SendGetRemainingTilesBatch(
+        Slice_Ptr_engine_GetRemainingTilesRequest([request])
```
bstrzelecki, line 112: Some wrapper methods would be nice to avoid such explicit casting

python_bindings/tests/engine_test.py
```diff
@@ -85,3 +91,102 @@ def test_game_engine_send_batch_returns_failures_when_communicator_closed(
...
+    )
+    resp.Err()  # the binding would raise if this returned an error
+    engine.Shutdown()
+
+    assert resp.Denominator == len(tiles)
```
bstrzelecki, line 117: What is the reason to go with such method of displaying probabilities, instead of simple floating point number?

python_bindings/tests/engine_test.py
```diff
@@ -85,3 +91,102 @@ def test_game_engine_send_batch_returns_failures_when_communicator_closed(
...
+    t1 = tiletemplates.MonasteryWithSingleRoad()
+    t2 = tiletemplates.RoadsTurn()
+    tiles = [t1, t2, t1]
+    tile_set = TileSet(
+        StartingTile=tiletemplates.SingleCityEdgeStraightRoads(),
+        Tiles=Slice_tiles_Tile(tiles),
+    )
+
+    engine = StartGameEngine(1, str(tmp_path))
+    game = engine.GenerateGame(tile_set)
```
bstrzelecki, line 108: Is there a way to clone a game state? Mostly to allow for simulating deeper branches of a specific game.

python_bindings/tests/engine_test.py
Jackenmen: The ability to represent a fraction exactly. If it's useless, I can divide each value instead.

python_bindings/tests/engine_test.py
Jackenmen: It's an annoyance with auto-generated bindings. There's no way to get gopy to generate methods that would do the casting on their own though I'm pretty sure gopy meant for it to just work and is just one of its bugs.
I should be able to write wrapper functions and assign them to the class's namespace to become methods since Python allows for that but it's going to be manual and slightly hacky.

python_bindings/tests/engine_test.py
Jackenmen: Nope. I can write a new request for cloning a game with the given ID that would return the new ID, should be simple to do.

bstrzelecki: Yeah, that would be nice since currently game needs to be replayed for every expansion of the game tree.

python_bindings/tests/engine_test.py
bstrzelecki: > I should be able to write wrapper functions and assign them to the class's namespace to become methods
This would allow us to become independent from auto-generated bindings. And also would yield a more stable API.

Jackenmen:
I realized that the linter was ignoring human-written wrappers for the bindings after

Jackenmen:
I realized that the linter was ignoring human-written wrappers for the bindings since before they were all machine-generated and it made sense to ignore the whole python_bindings folder. Now it doesn't so I updated the exclude rules for the linter to reflect that + fixed found issues.

Makefile
```diff
@@ -39,17 +39,18 @@ test-python: install-python
...
 	@echo "Running the linter..."
-	docker run \
+	docker run --rm \
 		-e "VALIDATE_ALL_CODEBASE=true" \
 		-e "DEFAULT_BRANCH=origin/main" \
 		-e "VALIDATE_GO=false" \
-		-e "FILTER_REGEX_EXCLUDE=.*python_bindings/.*" \
 		-e "VALIDATE_PYTHON_PYLINT=false" \
+		-e "LINTER_RULES_PATH=." \
+        -e "PYTHON_ISORT_CONFIG_FILE=pyproject.toml" \
```
bstrzelecki, line 48: wrong indentation

bstrzelecki (approved):

BognaLew (approved):

PR #62, opened 2024-08-05
Implement checking whether tile (including meeples) can be placed

### LINKED ISSUE #29: Implement a way to check if a meeple can be placed
Check if a meeple can be placed on a specific feature for a given tile placement.

Jackenmen:
Fixes #29
and updates all things (`GetLegalMovesFor`, `PlaceTile`) that should be using this functionality but couldn't because it didn't exist then.

### DIFF
--- a/pkg/game/board.go
+++ b/pkg/game/board.go
@@ -180,7 +180,8 @@ func (board *board) isPositionValid(tile elements.PlacedTile) bool {
 }
 
 func (board *board) CanBePlaced(tile elements.PlacedTile) bool {
-	return board.TileHasValidPlacement(elements.ToTile(tile)) &&
+	return board.isPositionValid(tile) &&
+		slices.Contains(board.placeablePositions, tile.Position) &&
 		// TODO: since `tile` can be user input, we need to check
 		// that legal number of meeples (i.e. just one) has been placed
 		board.cityCanBePlaced(tile) &&
--- a/pkg/game/board.go
+++ b/pkg/game/board.go
@@ -269,9 +269,11 @@ func (board *board) PlaceTile(tile elements.PlacedTile) (elements.ScoreReport, e
 			"Board's tiles capacity exceeded, logic error?",
 		)
 	}
-	// TODO for future tasks:
-	// - determine if the tile can placed at a given position,
-	//   or return ErrInvalidMove otherwise
+
+	if !board.CanBePlaced(tile) {
+		return elements.ScoreReport{}, elements.ErrInvalidPosition
+	}
+
 	setTiles := board.tileSet.Tiles
 	actualIndex := 1
 	for {
--- a/pkg/game/board.go
+++ b/pkg/game/board.go
@@ -172,11 +172,34 @@ func (board *board) isPositionValid(tile elements.PlacedTile) bool {
 	return true
 }
 
-//revive:disable-next-line:unused-parameter Until the TODO is finished.
 func (board *board) CanBePlaced(tile elements.PlacedTile) bool {
-	// TODO for future tasks:
-	// - implement generation of legal moves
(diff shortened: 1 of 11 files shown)


Jackenmen:
Ready for review with plans for a follow-up regarding Field API not allowing to expand on the checked tile: https://github.com/YetAnotherSpieskowcy/Carcassonne-Engine/issues/86

Kaszana102 (approved):
OK. No idea, what to do with Fields.

pkg/game/board.go
```diff
@@ -172,11 +186,119 @@ func (board *board) isPositionValid(tile elements.PlacedTile) bool {
...
+
+func (board *board) fieldCanBePlaced(tile elements.PlacedTile) bool {
+	// While placing a tile, the player can only put a single meeple on a field.
+	// This means we don't have to care about whether our field expands into
+	// a different feature on our tile - we will only expand the feature
+	// if it has a meeple and that only happens once.
+	for _, feat := range tile.Features {
+		if feat.FeatureType != feature.Field || feat.Meeple.Type == elements.NoneMeeple {
+			continue
+		}
```
Lenetis, line 235: - Since we only call fieldCanBePlaced on features with a meeple, checking if meeple exists is not necessary here
- Also, instead of iterating over all features on the tile and checking if the feature type is field, we could iterate over tile.GetFeaturesOfType(feature.Field) (although that also iterates through all features, so the end result would be slower in that case, so I'm not sure about this one)

pkg/game/board.go
```diff
@@ -172,11 +186,119 @@ func (board *board) isPositionValid(tile elements.PlacedTile) bool {
...
+func (board *board) monasteryCanBePlaced(_ elements.PlacedTile) bool {
+	// meeple can always be placed on a monastery
+	return true
+}
+
+func (board *board) roadCanBePlaced(checkedTile elements.PlacedTile) bool {
+	for _, feat := range checkedTile.Features {
+		if feat.FeatureType != feature.Road || feat.Meeple.Type == elements.NoneMeeple {
+			continue
+		}
```
Lenetis, line 267: Same thing as in fieldCanBePlaced

pkg/game/board.go
Jackenmen: > * Since we only call fieldCanBePlaced on features with a meeple, checking if meeple exists is not necessary here

It is necessary since only one of the field features will have a meeple and we don't know which without checking (though maybe that should just be passed to the function).

Lenetis: Right, I didn't think about that

pkg/game/board.go
```diff
@@ -117,11 +124,18 @@ func (board *board) TileHasValidPlacement(tile tiles.Tile) bool {
 	return false
 }
 
-func (board *board) GetLegalMovesFor(placement elements.PlacedTile) []elements.PlacedTile {
+func (board *board) GetLegalMovesFor(basePlacement elements.PlacedTile) []elements.PlacedTile {
```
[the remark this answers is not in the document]
Jackenmen: I updated the function to check whether the meeple can actually be placed there + documented the expected behaviour of the function.
To add to this comment, I want to mention that the function is primarily meant to be used with output from `GetTilePlacementFor()` and that's where the documented assumptions come from.

pkg/game/board.go
Jackenmen: I updated all of the functions to accept the feature and updated their implementation to use that fact.

pkg/game/board.go
Jackenmen: See https://github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pull/62#discussion_r1756906486

Lenetis (approved):

PR #52, opened 2024-07-22
2 game tests + automatic test job

Kaszana102:

PR #63, opened 2024-08-07
Final scoring

### LINKED ISSUE #64: A single tile in a road loop gets (incorrectly) counted twice
In the situation below, the blue meeple should be getting 4 points, not 5 points:
image

> When scoring a road, **each tile** of that road grants you **1 point**.
> [Source](https://wikicarpedia.com/car/Base_game#3._Scoring_a_road)

Here's a test case that reproduces this (based on the #63's branch though the issue is not related to what that PR is implementing):
https://github.com/Jackenmen/Carcassonne-Engine/commit/1a1473719508fa4176af6f600a08b66959801a0c

```
--- FAIL: TestFinalScoreMultipleRoads123 (0.00s)
    final_scoring_test.go:184: Player 1 final score incorrect. Expected 4, got: 5
0 :
1 :
2 :
3 :
FAIL
```

Kaszana102:
Fixes #64


bstrzelecki (approved):

pkg/game/final_scoring_test.go
```diff
@@ -0,0 +1,633 @@
...
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tiles/side"
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tiles/tiletemplates"
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tilesets"
+)
+
+/*
+  All tests contains the board visualization. Each tile is represented by 5x5 ascii characters
+*/
+
+func makeTurn(game *Game, t *testing.T, tilePosition position.Position, rotations uint, meeple elements.MeepleType, featureSide side.Side, featureType feature.Type) {
```
Jackenmen, line 21: 7 _positional_ arguments is a lot - it's hard to figure out the meaning of arguments from the call site alone.

Some of these arguments are closely related and could be grouped in a struct to improve readability - for example: `meeple`, `featureSide`, and `featureType` all refer to a meeple placement and are either set to None or to actual values.

With that said, personally, I found the `rotations` argument the most cryptic since at the call site it just looks like a magic number with an unknown meaning. Pre-rotating it in the tileset slice would both be more readable (since it would be easy to see that the number is passed to `.Rotate()` method) and allow you to remove another argument from this function. Alternatively, you could also turn all of the function's arguments into a single struct which would allow to specify the argument's name at the call site though I don't think it would be that beneficial, if the meeple placement arguments get grouped into a struct too.

pkg/game/final_scoring_test.go
```diff
@@ -0,0 +1,633 @@
...
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/stack"
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tiles"
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tiles/feature"
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tiles/side"
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tiles/tiletemplates"
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tilesets"
+)
+
+/*
+  All tests contains the board visualization. Each tile is represented by 5x5 ascii characters
```
Jackenmen, line 18: I looked up the format's description from the #52 PR (`pkg/end_tests/four_player_game_test/four_player_game_test.go`) but I want to note that it doesn't explain how `[`/`]` characters should be interpreted. They are related to monasteries but I'm not entirely sure what their placement means since I've seen variants where they occupy 3 rows and a case where it only occupies one row. Not a blocker for this PR but I would want some clarifications added in #52.

Jackenmen (changes requested):

pkg/game/board.go
```diff
@@ -439,11 +441,47 @@ func (board *board) ScoreRoads(placedTile elements.PlacedTile) elements.ScoreRep
...
 		checkedRoadSides |= road.Sides
 	}
 	return scoreReport
 }
+
+func (board *board) ScoreFinalMeeples() elements.ScoreReport {
+	meeplesReport := elements.NewScoreReport()
+
+	// score cities first (because they have their own manager)
+	meeplesReport.Join(board.cityManager.ScoreCities(true))
```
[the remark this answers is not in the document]
Kaszana102: but technically, do they need to me removed? Manager knows each city, and is able to score each once, so there is no risk of counting in multiple times.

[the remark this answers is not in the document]
Jackenmen: Technically no but I think it would be good to keep this consistent and avoid any unwanted surprises, in case it were to matter in some way in the future.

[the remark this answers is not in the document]
Kaszana102: done

pkg/game/final_scoring_test.go
Kaszana102: Is it now more acceptable? It may be a little bit better?

pkg/game/final_scoring_test.go
Kaszana102: I see in main board drawing in four_player_game_test I accidentaly drew monastery using single brackets[ ] insted of 3 pairs, but further in the test was drawn correctly. If you want I can change it to cross or another shape. I needed to somehow give it some surface to visualize that there is a meeple in/on it. That's why it not supposed to be drawn with single bracket pair

Jackenmen: > I see in main board drawing in four_player_game_test I accidentaly drew monastery using single brackets[ ] insted of 3 pairs, but further in the test was drawn correctly.

Ah, so it was just an unintentional inconsistency, that's fine - just make sure to fix that in #52 :) My comment about describing this in #52 still stands though ofc that's irrelevant to this PR.

> If you want I can change it to cross or another shape. I needed to somehow give it some surface to visualize that there is a meeple in/on it.

I'm mostly indifferent about the format right now since I have trouble visualizing the scenarios using these ASCII formats without recreating them in a fully graphical form anyway. I just noticed an inconsistency so wanted to know, if there's more to it. There are probably some things that could be done to make the ASCII format more readable but I don't know which part of it would need changing. I'd hold off on changing anything with it until this PR is merged, perhaps even until *after* #63 is merged.

pkg/game/final_scoring_test.go
Jackenmen: Definitely better! Lines are a bit lengthy but it's far from the only place in our codebase where that's the case :P

pkg/game/board.go
```diff
@@ -399,21 +415,26 @@ func (board *board) ScoreRoadCompletion(tile elements.PlacedTile, road feature.F
...
 
 	// check road in "right" direction
 	if !loopResult && rightSide != side.NoSide {
-		roadFinishedResult, scoreResult, meeplesResult, _, _ = board.CheckRoadInDirection(rightSide, tile)
+		roadFinishedResult, scoreResult, meeplesResult, _, _, finishedPosRight := board.CheckRoadInDirection(rightSide, tile)
 		score += scoreResult
 		roadFinished = roadFinished && roadFinishedResult
 		meeples = append(meeples, meeplesResult...)
+
+		// check if loop is on different tile
```
Jackenmen, line 430: I tried improving the explanation, what do you think?
```suggestion
		// Decrement the score to prevent counting the tile twice
		// when its road features (two different ones) are both the start
		// and the end of the road (as is the case for crossroads).
		//
		// Note that this is a different scenario than a literal loop (i.e. an actual circle
		// with no end or beginning) where we don't fall into this `if` branch at all
		// due to `loopResult` being `true`.
```

Jackenmen (changes requested):
Looks good functionally, I only left a few review comments about improvements to the code comments.

pkg/game/scoring_roads_test.go
```diff
@@ -111,8 +110,8 @@ func TestBoardScoreRoadLoopCrossroad(t *testing.T) {
 	}
 
 	// add meeple to first road
```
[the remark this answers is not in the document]
Kaszana102: fixed

pkg/game/board.go
```diff
@@ -316,31 +318,45 @@ param roadSide: always indicates only one cardinal direction!
...
 		}
 
-		score++
-
 		// Get road feature
 		road = tile.GetPlacedFeatureAtSide(roadSide, feature.Road)
 
+		// check if loop
+		if tile.Position == startTile.Position {
+			// before finishing check for meeple on potentialy other side of crossroad
```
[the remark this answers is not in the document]
Kaszana102: quite long, but ok

Jackenmen (approved):

PR #68, opened 2024-08-20
Positions and "is placed" bits in binary tiles

Lenetis:


pkg/tiles/binarytiles/binary_tile.go
```diff
@@ -192,14 +216,30 @@ func (binaryTile *BinaryTile) addOrthogonalFeature(feature elements.PlacedFeatur
...
 }
 
+// Sets the position X and position Y bits in the binary tile
+func (binaryTile *BinaryTile) addPosition(position position.Position) {
+	if position.X() > 127 || position.Y() > 127 || position.X() < -128 || position.Y() < -128 {
+		panic(fmt.Sprintf("position %#v out of range for binary tile. Allowed range: [-128, 127]", position))
+	}
+	var tmpBinaryTile BinaryTile
+	tmpBinaryTile |= BinaryTile(uint8(position.X() + 128))
+	tmpBinaryTile <<= 8
```
bstrzelecki, line 237: Use actual size of uint8 instead of hard-coded number

pkg/tiles/binarytiles/binary_tile.go
```diff
@@ -192,14 +216,30 @@ func (binaryTile *BinaryTile) addOrthogonalFeature(feature elements.PlacedFeatur
...
+	tmpBinaryTile |= BinaryTile(uint8(position.Y() + 128))
+	tmpBinaryTile <<= 48
```
bstrzelecki, line 239: We should use calculated offset instead of hard-coded number

pkg/tiles/binarytiles/binary_tile.go
```diff
@@ -25,11 +32,14 @@ type BinaryTile int64
 //  - next four meeple bits are the corners (same as with fields)
 //  - the last meeple bit is the center
 //  - the owner bits is just one-hot-encoded player ID. (ID(1) = 00...001, ID(2) = 00...010, etc.)
+//  - is placed bit is always 1 on all placed tiles, and 0 on the non-placed tiles
+//  - position bits are 8-bit reptesentations of tile position + 128, so that there are no negative numbers
```
bstrzelecki, line 36: What is the reasoning behind converting to non-negative numbers?

Lenetis: I had some problems with converting positions from int8 to uint64 (a.k.a. to BinaryTile) and writing test cases for that, but I figured it out now

BognaLew (approved):

PR #69, opened 2024-08-20
Fixed isPositionValid

### LINKED ISSUE #66: `board.isPositionValid()` does not handle fields overlapping with other features correctly
The `MonasteryWithSingleRoad()` tile has a road on its bottom side and a field feature on all sides:
image
https://github.com/YetAnotherSpieskowcy/Carcassonne-Engine/blob/790609dd6e9edac0353ad06b70101ecc5ed9e6d1/pkg/tiles/tiletemplates/tile_templates.go#L34-L63

A case where sides overlap appears to not have been considered when implementing `board.isPositionValid()`/`board.testSide()`.

Regression test:
https://github.com/YetAnotherSpieskowcy/Carcassonne-Engine/commit/ec151a6517500e9c17474c2fc0eb444c5dccfeeb

Graphical example of the regression test:
image

Lenetis:
Fixes #66

### DIFF
--- a/pkg/game/board.go
+++ b/pkg/game/board.go
@@ -109,9 +109,16 @@ func (board *board) GetLegalMovesFor(tile elements.PlacedTile) []elements.Placed
 	return []elements.PlacedTile{}
 }
 
-// Returns true if the tile placement position is valid, i.e. if all existing neighbouring tiles have matching features.
-// (for example, city feature directly neighbouring road or field is not valid)
-// Does not take meeples into account.
+/*
+Returns true if the tile placement position is valid, i.e. if all existing neighbouring tiles have matching features.
+(for example, city feature directly neighbouring road or field is not valid)
+
+Only checks the validity of the tile placement based on the tile features and their neighbors. It does not take into account:
+- The placement of meeples
+- Whether the tile is being placed on an already occupied position
+- Whether the tile is not neighboring any tiles
+- Whether the tile has already been placed somewhere else on the board
+*/
 func (board *board) isPositionValid(tile elements.PlacedTile) bool {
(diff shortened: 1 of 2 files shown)


Jackenmen (dismissed):
I found an issue:
image

I pushed a test case that you can reproduce this with.

pkg/game/board.go
```diff
@@ -139,14 +109,35 @@ func (board *board) GetLegalMovesFor(tile elements.PlacedTile) []elements.Placed
...
+				}
+			}
+
+		}
+	}
+	// roads are checked separately because when placing a tile with fields neighbouring a tile with roads,
+	//  features checked from this tile to the neighbour are still matching (every field has a corresponding field on the neighbour tile),
+	//  but roads from the other tile might still not have a corresponding road in this one, despite the previous check returning true
+	//
+	// TODO: figure out a cleaner way to do this
```
Jackenmen, line 132: To be honest, once I understood why this was needed, I no longer considered it unclean. The original comment focuses on how this is technically specific to roads but that's really just an optimization strategy - the general idea is that due to features potentially having overlaps, it is not enough to check that our features have counterparts since the neighbour could also have **another** feature on the side that we checked.

On a high-level, we're doing:
```py
# input: tile
for side in EDGE_SIDES:
    neighbour = board.neighbour(tile, side)
    if neighbour and tile.features_on(side) != neighbour.features_on(side.mirror()):
        return False

return True
```
It's just that our current APIs don't have a way to express this without affecting performance negatively.

Jackenmen (changes requested):

Jackenmen (approved):

Kaszana102 (approved):

PR #74, opened 2024-08-26
performance tests

Kaszana102:
created simple performance tests for all features. Not sure why monastery long test is much worse, than other test.


pkg/game/performancetests/cities_test.go
```diff
@@ -0,0 +1,81 @@
+//go:build performanceTest
```
bstrzelecki, line 1: this should no longer be necessary

.github/workflows/performance.yml
```diff
@@ -0,0 +1,30 @@
...
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: "${{ env.GO_VERSION }}"
+      - name: Install gocover-cobertura
+        run: |
+          go get github.com/boumenot/gocover-cobertura && go install github.com/boumenot/gocover-cobertura
```
bstrzelecki, line 25: we don't need coverage dependencies here

pkg/game/performancetests/cities_test.go
Kaszana102: removed

.github/workflows/performance.yml
```diff
@@ -0,0 +1,27 @@
...
+  workflow_dispatch:
+
+env:
+  GO_VERSION: '1.22.1'
+  PYTHON_VERSION: '3.12'
+  BUILT_WHEELS_DIR: built_wheels
+
+permissions: read-all
+jobs:
+  go-tests:
```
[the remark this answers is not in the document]
Kaszana102: renamed

BognaLew (approved):

.github/workflows/performance.yml
```diff
@@ -0,0 +1,27 @@
+name: Performance tests
+
+on:
```
bstrzelecki, line 3: We might want to run benchmarks for every push to the main branch, so we have a reference point

Kaszana102: changed

.github/workflows/performance.yml
```diff
@@ -0,0 +1,31 @@
...
+  push:
+    branches:
+      - main
+  pull_request:
+    branches: main
```
bstrzelecki, line 8:
```suggestion
```
Run only for every push to main branch

bstrzelecki (approved):

PR #85, opened 2024-09-04
Recover from panics in GameEngine requests

### LINKED ISSUE #82: Python bindings - recover from panics
We do not want to bring down the whole runtime due to a panic somewhere in our implementation - implement recover() pattern in the game engine to catch any panic that may arise: https://stackoverflow.com/questions/25025467/catching-panics-in-golang

Jackenmen:
Fixes #82 

Comes with a refactor of `sendBatch` to a new class as the original implementation which became rather lengthy and not particularly easy to read.

Diff from before the refactor:
https://github.com/YetAnotherSpieskowcy/Carcassonne-Engine/compare/f07e719526a5b63ebe84f1c1f0132a58f59ed736

### DIFF
--- a/pkg/engine/base_structures.go
+++ b/pkg/engine/base_structures.go
@@ -77,3 +77,12 @@ func (resp *BaseResponse) Err() error {
 type SyncResponse struct {
 	BaseResponse
 }
+
+func (resp *SyncResponse) canRemoveGame() bool {
+	err := resp.Err()
+	if err == nil {
+		return false
+	}
+	_, panicOccured := err.(*ExecutionPanicError)
+	return panicOccured
+}
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path"
 	"runtime/debug"
+	"strings"
 	"sync"
 
 	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/game"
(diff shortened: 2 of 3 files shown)


pkg/engine/base_structures.go
```diff
@@ -77,3 +77,12 @@ func (resp *BaseResponse) Err() error {
...
 }
+
+func (resp *SyncResponse) canRemoveGame() bool {
+	err := resp.Err()
+	if err == nil {
+		return false
+	}
+	_, panicOccured := err.(*ExecutionPanicError)
+	return panicOccured
+}
```
Lenetis, line 88: Can we only remove games that have errors? I'm not 100% sure how this works

pkg/engine/engine.go
```diff
@@ -308,145 +345,200 @@ func (engine *GameEngine) SendGetLegalMovesBatch(concreteRequests []*GetLegalMov
 //
 // If a request fails before reaching a worker, a `SyncResponse` type will be
```
Lenetis, line 346: the 'be' is repeated in the next line
```suggestion
// If a request fails before reaching a worker, a `SyncResponse` type will
```

Lenetis (approved):

pkg/engine/base_structures.go
Jackenmen: See the interface definition:
```go
// Responses implementing this interface may indicate to the sender
// that the the game with `GameID()` should be removed from the engine.
type ResponseGameRemovable interface {
    canRemoveGame() bool
}
```

Basically, `canRemoveGame()` is a way to inform the engine that there is no point in the user making any more requests on that game.
For `PlayTurnResponse` that is the case when there are no more tiles to place as that indicates that the game is finished.
With the changes introduced in this PR, this is now also the case when a panic occurs during execution since it can leave the game in an inconsistent state.

bstrzelecki (approved):

PR #89, opened 2024-09-09
Fix expansion to another field feature when validating placement

### LINKED ISSUE #86: Field instance cannot expand to another feature on the checked tile, if it's not on the board
### Description
This TODO needs to be addressed:
https://github.com/YetAnotherSpieskowcy/Carcassonne-Engine/blob/a711a81f37b51efd849b63710c001a60ab8971fb/pkg/game/board.go#L240-L243

### Graphical example 
image

### Regression test
Already present in the affected branch (#62), you'll just need to remove the `t.Skip` to have it run:
https://github.com/YetAnotherSpieskowcy/Carcassonne-Engine/blob/a711a81f37b51efd849b63710c001a60ab8971fb/pkg/game/board_test.go#L137-L177

### Test output
```
$ go test -count=1 ./pkg/game -run TestBoardFieldCanBePlacedReturnsFalseWhenExpandToFieldWithMeepleHappensOverAnotherField
--- FAIL: TestBoardFieldCanBePlacedReturnsFalseWhenExpandToFieldWithMeepleHappensOverAnotherField (0.00s)
    board_test.go:173: expected false, got true instead
FAIL
FAIL	github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/game	0.002s
FAIL
```

Lenetis:
Fixes #86

### DIFF
--- a/pkg/game/board.go
+++ b/pkg/game/board.go
@@ -246,12 +246,6 @@ func (board *board) fieldCanBePlaced(tile elements.PlacedTile, feat elements.Pla
 	// This means we don't have to care about whether our field expands into
 	// a different feature on our tile - we will only expand the feature
 	// if it has a meeple and that only happens once.
-
-	// unset the Meeple in our copy of the feature
-	feat.Meeple = elements.Meeple{}
-
-	// TODO: Field instance cannot expand to another feature on the tile checked by
-	// this method since it's not part of the board. See GH-86.
 	field := field.New(feat, tile)
 
 	return field.IsFieldValid(board, 1)
--- a/pkg/game/board.go
+++ b/pkg/game/board.go
@@ -252,14 +252,9 @@ func (board *board) fieldCanBePlaced(tile elements.PlacedTile, feat elements.Pla
 
 	// TODO: Field instance cannot expand to another feature on the tile checked by
 	// this method since it's not part of the board. See GH-86.
-	field := field.New(feat, tile.Position)
-	field.Expand(board, board.cityManager)
+	field := field.New(feat, tile)
 
-	// score report function checks the whole field for placed meeples
-	// and reports any meeples that would be returned which we can use here
-	scoreReport := field.GetScoreReport()
-
-	return len(scoreReport.ReturnedMeeples) == 0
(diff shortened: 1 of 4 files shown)


pkg/game/board.go
```diff
@@ -252,14 +252,9 @@ func (board *board) fieldCanBePlaced(tile elements.PlacedTile, feat elements.Pla
 
 	// TODO: Field instance cannot expand to another feature on the tile checked by
```
bstrzelecki, line 253: Is this comment still necessary?

pkg/game/field/field.go
```diff
@@ -119,16 +132,51 @@ func (field *Field) Expand(board elements.Board, cityManager city.Manager) {
 	field.features = newFeatures
 }
 
+// Returns true if the field has lesser or equal number of meeples than maxMeepleCount, and false otherwise
```
Jackenmen, line 135: While this description of the function is technically true, it does not necessarily make it clear how this is useful, i.e. it can be used to check whether a meeple could be placed on the given field.
With that use case in mind, the way to achieve this should probably also be documented since it's not necessarily clear what `maxMeepleCount` should be set to - this actually depends on whether the tile has any meeple on a field feature (always the case for the only usage - i.e. `fieldCanBePlaced` - which is why it needs to be passed with `1`).

pkg/game/field/field.go
```diff
@@ -119,16 +132,51 @@ func (field *Field) Expand(board elements.Board, cityManager city.Manager) {
...
+func (field Field) IsFieldValid(board elements.Board, maxMeepleCount int) bool {
```
Jackenmen, line 136: a) it should be documented that this method cannot be used twice since it alters the `field`'s state (maps are pointers so they point to the same map on copy)
b) it would be easier to realise that this can modify the field if it were using a pointer receiver

Jackenmen (changes requested):

pkg/game/field/field.go
Lenetis: changed so that it doesn't modify the original field.features, and instead works on a copy

bstrzelecki (approved):

PR #99, opened 2024-09-16
Finalize game during PlayTurnRequest if it's the last turn

Jackenmen:


Kaszana102 (approved):
game.Finalize() in each turn creeps me out, but it looks ok.

PR #101, opened 2024-09-17
Update game's current player only if tile was successfully placed

Kaszana102:


pkg/game/game.go
```diff
@@ -222,14 +222,15 @@ func (game *Game) PlayTurn(move elements.PlacedTile) error {
...
 	player := game.CurrentPlayer()
-	defer func() { game.currentPlayer = (game.currentPlayer + 1) % game.PlayerCount() }()
 
 	// In the class diagram, the `scoreReport` would be returned by
 	// separate `CheckCompleted()` method but it's been abstracted by PlaceTile instead.
 	scoreReport, err := player.PlaceTile(game.board, move)
 	if err != nil {
 		return err
 	}
+	defer func() { game.currentPlayer = (game.currentPlayer + 1) % game.PlayerCount() }() // change player after placing tile, in case of incorrect turn
```
Jackenmen, line 232: This doesn't really need to be deferred. Looking at [the original code that introduced this](https://github.com/Jackenmen/Carcassonne-Engine/commit/721ff66469247683faf5e4e3d8914f761ad61fbd#diff-bed4390b3de58519900c0a6f7e7b994b2af974f618d3c774f2b5afa73f2d37d2R112), I imagine it was deferred because the code used to refer to the `game.currentPlayer` and so that had to stay the same even after the tile was placed. This is no longer the case and as such this can be simplified:
```suggestion
	game.currentPlayer = (game.currentPlayer + 1) % game.PlayerCount()
```

Jackenmen (changes requested):

Jackenmen (approved):

BognaLew (approved):

PR #106, opened 2024-09-18
Ensure input to Game.PlayTurn is fully copied

### LINKED ISSUE #105: GetLegalMovesRequest seems not returning all moves
https://github.com/YetAnotherSpieskowcy/Carcassonne-Engine/blob/4e58c78935e4514669664661a4e06d5999e7e81e/pkg/engine/engine_test.go#L676-L736

There is no valid turn where the tile is placed above the starting tile, and player can place a meeple on the city. Notice that the city will be closed as the tile will be placed.
GetLegalMoves returns 5 possible moves for tile in position (0,1), however two are empty (without meeple). It correctly creates move with meeple on a:
- bottom field
- upper field
- road
The request should return 5 moves for that position, but one  move is wrong.

In total request returns 30 moves, but we are interested in a tile on certain position.

Jackenmen:
Fixes #105 

The problem in the issue is triggered by `GetLegalMovesFor.execute()` passing `PlacedTile` to `Game.PlayTurn()`, expecting it to not be mutated. This is not actually the case though as the underlying slice is never copied. This PR fixes that. Since the pattern of copying `Features` can be seen in a few places in the codebase already, I went ahead and replaced them with the new `PlacedTile.DeepClone()` method call.

### DIFF
--- a/pkg/game/board.go
+++ b/pkg/game/board.go
@@ -141,8 +141,7 @@ func (board *board) GetLegalMovesFor(basePlacement elements.PlacedTile) []elemen
 
 	for i := range basePlacement.Features {
 		for _, meepleType := range meepleTypes {
-			placement := basePlacement
-			placement.Features = slices.Clone(basePlacement.Features)
+			placement := basePlacement.DeepClone()
 			placement.Features[i].Meeple = elements.Meeple{Type: meepleType}
 			feat := placement.Features[i]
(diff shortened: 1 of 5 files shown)


Kaszana102 (approved):
yop, it fixes the problem

BognaLew (approved):

PR #118, opened 2024-09-23
Fix PlacedTile immutability assumption in board.DeepClone()

Jackenmen:
Before meeple removal was added (#55), `PlacedTile` was considered immutable. This already needed to be fixed in `Game.PlayTurn()` (#106) but today I found another place where this assumption was made but is no longer correct.

Ideally, I would actually prefer to remove the mutation of `PlacedTile` by updating `RemoveMeeple()` to modify a clone of the tile but that is not as easy to do - we'd need to update both `board.tiles` and `board.tilesMap` (the former requires search) as well as ensure that any code looking up the tile once (probably just `board.ScoreMeeples()`) is changed to look the tile up again *if* it performed meeple removal. In short: not worth it.

The key part is https://github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pull/118/commits/571610c23c82ba731c7779520d5dbde47cb3140f
The other commits are adding tests and improving existing game/board split to make those tests viable - having to remove the meeples on the game side of things is a weird design choice.

### DIFF
--- a/pkg/game/board.go
+++ b/pkg/game/board.go
@@ -306,6 +306,11 @@ func (board *board) roadCanBePlaced(checkedTile elements.PlacedTile, checkedRoad
 	return true
 }
 
+// Add a tile to the board and propagate feature completion
+// to other tiles on the board (including meeple removal).
+//
+// Anything not managed by the board, such as players, will need to be updated
+// by the caller.
 func (board *board) PlaceTile(tile elements.PlacedTile) (elements.ScoreReport, error) {
 	err := board.addTileToBoard(tile)
 	if err != nil {
@@ -314,6 +319,8 @@ func (board *board) PlaceTile(tile elements.PlacedTile) (elements.ScoreReport, e
 	return board.checkCompleted(tile), nil
 }
 
+// Add a tile to the board without propagating feature completion to
+// other tiles on the board or removing meeples.
 func (board *board) addTileToBoard(tile elements.PlacedTile) error {
 	if board.TileCount() == cap(board.tiles) {
 		return errors.New("Board's tiles capacity exceeded, logic error?")
--- a/pkg/game/board.go
+++ b/pkg/game/board.go
@@ -337,8 +337,8 @@ func (board *board) PlaceTile(tile elements.PlacedTile) (elements.ScoreReport, e
 	board.updateValidPlacements(tile)
 	board.tiles[actualIndex] = tile
 	board.tilesMap[tile.Position] = tile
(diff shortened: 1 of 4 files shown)


pkg/game/board.go
```diff
@@ -73,9 +72,23 @@ func NewBoard(tileSet tilesets.TileSet) elements.Board {
...
+		tile = tile.DeepClone()
+		tilesMap[pos] = tile
+	}
+	board.tilesMap = tilesMap
+
+	tiles[0] = board.tiles[0]
+	for i, tile := range board.tiles {
+		if tile.Position.X() == 0 && tile.Position.Y() == 0 {
+			continue
+		}
```
Kaszana102, line 88: Is it possible to simply check if i==0, or before loop create shallow copy, without first tile? To ignore checking if starting tile in every iteration.

pkg/game/board.go
```diff
@@ -358,20 +375,20 @@ func (board *board) updateValidPlacements(tile elements.PlacedTile) {
...
 
-func (board *board) checkCompleted(
-	//revive:disable-next-line:unused-parameter Until the TODO is finished.
-	tile elements.PlacedTile,
-) (elements.ScoreReport, error) {
-	// TODO for future tasks:
-	// - identify all completed features
-	// - resolve control of the completed features
-	// - award points
+func (board *board) checkCompleted(tile elements.PlacedTile) elements.ScoreReport {
```
Kaszana102, line 378: Please add comment to this function description, that it removes meeples, but meeples count still needs to be updated using the functions result. It may not obvious, due to moving the functionality here.

Kaszana102 (changes requested):

pkg/game/board.go
Jackenmen: That's, surprisingly, not the point of this check - the `Tiles` slice is actually of fixed size (was meant to be used as a vector for the agent) and not all tiles in it are real. This is one way of checking it though there are probably some simpler ones - I could just check if `tile.Features == nil`. Either way, I should probably add a comment for this.

pkg/game/board.go
Jackenmen: I added a basic description of the `PlaceTile()` and `addTileToBoard()` methods with a note about this. I do feel it should be obvious since the board has no knowledge of players (it does not encompass them - the `Game` does) but I suppose it doesn't hurt.

Kaszana102 (approved):

BognaLew (approved):

PR #126, opened 2024-09-27
Fix checking other side of road with noside Side on start tile in road scoring

Kaszana102:


Kaszana102:
Main issue was that every side has NoSide side, so GetPlacedFeatureAtSide will return first feature with correct type, when called with NoSide value. Added check before calling it, to prevent it

pkg/game/board.go
```diff
@@ -570,7 +570,11 @@ func (board *board) scoreRoadCompletion(tile elements.PlacedTile, road feature.F
 
 	// check meeples on start tile
 	var roadLeft = tile.GetPlacedFeatureAtSide(leftSide, feature.Road)
-	var roadRight = tile.GetPlacedFeatureAtSide(rightSide, feature.Road)
+	var roadRight *elements.PlacedFeature
+	if rightSide != side.NoSide {
```
Jackenmen, line 574:
```suggestion
	// If a road doesn't connect two sides (i.e. ends in the centre)
	// then it will not have a "right" side and this variable will be 0
	if rightSide != side.NoSide {
```

Jackenmen (commented):
Looks good, I'd just like a comment added to clarify the reason for the added `if` - see the comment below.

Jackenmen (approved):

BognaLew (approved):

PR #128, opened 2024-09-27
Requests benchmark tests

Kaszana102:
The only thing left, is converting action to on click only, i will convert to it, when approved


bstrzelecki:
Those benchmark will spend more time managing timers than actually running engine code. There is too little processing between time measurements, that it will not yield any meaningful results. And also they are testing only small sliver of possible requests that won't be representative of actual use of the engine.

Kaszana102:
> Those benchmark will spend more time managing timers than actually running engine code. There is too little processing between time measurements, that it will not yield any meaningful results. And also they are testing only small sliver of possible requests that won't be representative of actual use of the engine.

I'm sorry, that engine needs setting up? Or even that maketurn request mutates the game, so the game needs to be copied beforehand? Each request is tested in 3 three different timestamps of the same game, technicaly more are possible, but I believe those 3 show the complexity tendency. What do you mean by "small sliver of possible requests"? Each reuqest is tested. Each test uses the same game setup, which allows comparing all request in same scenario in different stage of a game. Making up any new scenario game is quite time consuming. Guess who was only making those?

Please provide any suggestions or fixes, which would please you. Your comment doesn't suggest any solution.

bstrzelecki:
Have you at the very least checked if the code you are pushing is actually useful?
> I'm sorry, that engine needs setting up? Or even that maketurn request mutates the game, so the game needs to be copied beforehand?

I'm not getting were you got that was an issue, mine problem was that while running benchmark almost half of the execution time is spent in starting and stopping timers, not actually running engine code. This mostly due to processing time of single request is too short to be reliably measured.

> Each request is tested in 3 three different timestamps of the same game, technicaly more are possible, but I believe those 3 show the complexity tendency. What do you mean by "small sliver of possible requests"? Each reuqest is tested. Each test uses the same game setup, which allows comparing all request in same scenario in different stage of a game. 

In this case you are measuring execution time of performing few particular requests in few particular circumstances which is in no way comparable to actual use case of the engine and thus rendering results of these benchmarks meaningless.

> Making up any new scenario game is quite time consuming. Guess who was only making those?

No one asked you to come up with specific game scenarios, utilizing many random(but reproducible) game is in every way sufficient for the purpose of the benchmarks.

> Please provide any suggestions or fixes, which would please you. Your comment doesn't suggest any solution.

Measuring processing time of requests made during full game (e.g. [this test](https://github.com/YetAnotherSpieskowcy/Carcassonne-Engine/blob/5c15ff011aa7d58299a63c5d018ddc9d82d467d7/python_bindings/tests/engine_test.py#L25) but in go) would be more representative sample. Also starting multiple games at the same time and then sending requests from those games in a single batch would solve the problem with timers mentioned earlier.

Jackenmen:
> e.g. [this test](https://github.com/YetAnotherSpieskowcy/Carcassonne-Engine/blob/5c15ff011aa7d58299a63c5d018ddc9d82d467d7/python_bindings/tests/engine_test.py#L25) but in go

Just as an FYI for potentially saving time, this has a Go counter-part which could be useful:
https://github.com/YetAnotherSpieskowcy/Carcassonne-Engine/blob/5c15ff011aa7d58299a63c5d018ddc9d82d467d7/pkg/engine/engine_test.go#L54

(+ the fuzzer also has a similar code https://github.com/YetAnotherSpieskowcy/Carcassonne-Engine/commit/3b7c266f7298a61d24ba55ac000791a39011d2ba)

pkg/engine/request_performance_tests/setup_engine.go
Lenetis, line 1: Please add some short comments explaining what each function in this file does. (PlayTurns specifically is a bit complicated to me - it isn't used anywhere and it returns a pointer to the engine and a serialized game. I think I understand what it does, but what is it used for?)

pkg/tilesets/tilesets.go
```diff
@@ -141,3 +141,37 @@ func StandardTileSet() TileSet { //nolint:gocyclo // shallow loops for adding ti
 		Tiles:        tiles,
 	}
 }
+
+func EveryTileOnceTileSet() TileSet { //nolint:gocyclo // shallow loops for adding tiles
+	var tiles = []tiles.Tile{}
+
+	tiles = append(tiles, tiletemplates.MonasteryWithoutRoads()) // 1
+	tiles = append(tiles, tiletemplates.MonasteryWithSingleRoad().Rotate(1))
+	tiles = append(tiles, tiletemplates.StraightRoads()) // 3
```
Lenetis, line 150: Since there aren't any loops here (like in the standard tileset), this could be simplified as:
```go
var tiles = []tiles.Tile{
    tiletemplates.MonasteryWithoutRoads(),
    tiletemplates.MonasteryWithSingleRoad().Rotate(1),
    tiletemplates.StraightRoads(),
    ...
}
```

Kaszana102:
Is it better now?

pkg/engine/request_performance_tests/play_game.go
```diff
@@ -0,0 +1,75 @@
...
+	"testing"
+
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/engine"
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tilesets"
+)
+
+type TestedRequest func(games *[]engine.SerializedGameWithID, eng *engine.GameEngine, b *testing.B)
+
+func PlayGame(gameCount int, b *testing.B, testedRequest TestedRequest) {
+	b.StopTimer()
```
Lenetis, line 13: Why is there a StopTimer here?

pkg/engine/request_performance_tests/play_game.go
```diff
@@ -0,0 +1,75 @@
...
+		if err != nil {
+			b.Fatal()
+		}
+		games = append(games, game)
+	}
+
+	// for each turn
+	for range len(tilesets.StandardTileSet().Tiles) {
+
+		// at start before makiny any turn
```
Lenetis, line 32:
```suggestion
		// at the start before making any turn
```

pkg/engine/request_performance_tests/play_game.go
```diff
@@ -0,0 +1,75 @@
...
+			)
+		}
+		playTurnResp := eng.SendPlayTurnBatch(makeTurnRequests)
+
+		// update games
+		for i := range gameCount {
+			games[i].Game = playTurnResp[i].Game
+			if playTurnResp[i].Err() != nil {
+				b.Fatalf("%#v play turn failed. Reason: %#v", i, playTurnResp[i].Err().Error())
+			}
```
Lenetis, line 72: I'd reorder these two parts - no reason to assign the game if it has an error and will fail in the next line
```suggestion
			if playTurnResp[i].Err() != nil {
				b.Fatalf("%#v play turn failed. Reason: %#v", i, playTurnResp[i].Err().Error())
			}
			games[i].Game = playTurnResp[i].Game
```

pkg/engine/request_performance_tests/play_game.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"testing"
+
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/engine"
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/tilesets"
+)
+
+type TestedRequest func(games *[]engine.SerializedGameWithID, eng *engine.GameEngine, b *testing.B)
+
+func PlayGame(gameCount int, b *testing.B, testedRequest TestedRequest) {
```
Lenetis, line 12: In all other functions `b *testing.B` is the last argument. For consistency, I'd advice to do it like so here as well

pkg/engine/request_performance_tests/play_game.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	// for each turn
+	for range len(tilesets.StandardTileSet().Tiles) {
+
+		// at start before makiny any turn
+		// test desired requests
+		for range b.N {
+			testedRequest(&games, eng, b)
+		}
```
Lenetis, line 36: Asking just to make sure - this is here so that we can test the requests in a "real" environment, during an actual game, instead of on an engine that has just started, right?

(At first I wasn't sure why do we need the whole PlayGame method, when only this line is the one that actually measures anything)

Lenetis (changes requested):

pkg/engine/request_performance_tests/clone_game_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+	"github.com/YetAnotherSpieskowcy/Carcassonne-Engine/pkg/engine"
+)
+
+func SendCloneGameRequest(games *[]engine.SerializedGameWithID, eng *engine.GameEngine, b *testing.B) {
+	var ids []int
+	var err error
+	b.StartTimer()
+	for _, game := range *games {
+		ids, err = eng.CloneGame(game.ID, 10)
```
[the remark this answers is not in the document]
Kaszana102: fixed

pkg/engine/request_performance_tests/play_game.go
Kaszana102: because we want to measure only one of the specified request. We want to ignore any other delays sucha as: game set up, playing turn or getting legal moves. We want to measure how fast are request when called at the begining of the turn. That's why there's call  testedRequest(&games, eng, b), so the inside takes care of measurements. You can that inside tested function are Start and Stop timers

pkg/engine/request_performance_tests/play_game.go
Kaszana102: fixed

pkg/engine/request_performance_tests/play_game.go
Kaszana102: small nitpick but ok

pkg/engine/request_performance_tests/play_game.go
Kaszana102: consistenced

pkg/engine/request_performance_tests/play_game.go
Kaszana102: We want to measure the average time of a certain request during whole game, so we need to play it, and measure the request each turn. We test multiple different games at once, to have more reliable measurements. Moreover we check all turns, instead of single case, witch would some previous special setup.

pkg/engine/request_performance_tests/play_game.go
Lenetis: Ok, but what specific timer are we stopping here? If I understand correctly, there should be no timer started in this place - each tested request starts and stops timers within itself, and no more timers are started besides that. (ctrl+f finds 5 occurences of StartTimer and 6 of StopTimer). So I still don't understand

Kaszana102: timer literaly starts at the beginning of a test by default.

Lenetis: Ooo, now everything is clear, thanks

Lenetis (approved):

.github/workflows/requests_performance .yml
bstrzelecki, line 1: What is the reasoning for this being a separate job from the previous benchmarks?

pkg/engine/engine_test.go
```diff
@@ -587,3 +587,14 @@ func TestGameEngineSendBatchReturnsExecutionPanicErrorOnPanicEverywhere(t *testi
 	}
 	engine.Close()
 }
+
+func TestGenerateSeededGame(t *testing.T) {
```
bstrzelecki, line 591: What does this test?

.github/workflows/requests_performance .yml
Kaszana102: as i said in the pull request first comment, i will revert to onclick

bstrzelecki: What I meant is why it is not the part of [this](https://github.com/YetAnotherSpieskowcy/Carcassonne-Engine/blob/main/.github%2Fworkflows%2Fperformance.yml)

Kaszana102: combined

pkg/engine/engine_test.go
Kaszana102: Now compares the game's deck and seeded deck

.github/workflows/performance.yml
```diff
@@ -26,4 +26,20 @@ jobs:
       - name: Set up gotestfmt
         uses: GoTestTools/gotestfmt-action@v2            
       - name: Run tests
-        run: go test -v ./pkg/game/performancetests -bench=. -run=xxx -benchmem    
\ No newline at end of file
+        run: go test -v ./pkg/game/performancetests -bench=. -run=xxx -benchmem    
+        
+  request-performance-tests:
```
bstrzelecki, line 31: Okay, but there is no benefit compared to the previous version since we are setting up go and building our project twice anyways

bstrzelecki: It can be incorporated into a singe benchmark run

bstrzelecki (approved):

PR #132, opened 2024-09-30
Allow GameEngine to not create logs

Jackenmen:


pkg/engine/requests_test.go
```diff
@@ -471,3 +471,175 @@ func TestGameEngineSendGetMidGameScoreBatchAtMidGameReturnsExpectedScores(t *tes
...
+	}
+}
+
+func TestGetLegalMovesAndPlayItWithPlayTurnRequestWithNoLogger(t *testing.T) {
+	engine, err := StartGameEngine(4, "")
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+
+	game, err := engine.GenerateGame(tilesets.StandardTileSet()) // Preferably change for GenerateSeededGame when benchmarks are merged
```
Lenetis, line 516: I'd suggest adding a `TODO` in this comment, so that it's easier to find and we don't forget it in the future.
Same thing with the comment in line 544.

Lenetis (approved):

pkg/engine/requests_test.go
Kaszana102: TODOed

bstrzelecki (approved):

PR #143, opened 2024-10-17
Fix performance tests GH Actions workflow

Jackenmen:


Kaszana102 (approved):
I'm glad that you've fixed it before i returned home, but i still wonder why it doesn't show any error in terminal, when i try to do it in the wrong way :/

bstrzelecki (approved):

PR #144, opened 2024-10-20
Changed int16 to int8 in positions

Lenetis:


Jackenmen (dismissed):

Kaszana102:
I believe benchmark tests will fail (the long features), where there are 1000 tiles placed in a row. You should check it

Kaszana102 (approved):

pkg/game/performancetests/quick_game.go
```diff
@@ -13,8 +13,8 @@ import (
 )
 
 /*
-Quick function for playing a simple game in a straight line.
-Arugment playGame is used for cases when empty game needs to be measured.
+Quick function for playing a simple game
+The tiles are placed in a straight line from x=0 to x=125. After that a new row is started (x=0, y=1) and so on
```
Jackenmen, line 17: Hmm, this will probably cause some test cases to have a different performance characteristic than a long line would in the case of a road where a function has to visit all parts of a single road since now that road is actually going to be ~40 smaller roads. Using road turns on edges would keep the behaviour similar but would require not using this function.

Fields and cities are unaffected since they can extend in any direction while monasteries are *probably* unaffected (they would be affected if a meeple were placed on the tile since then we would see that most of the monasteries would actually get to be scored but the tests do not place meeples as far as I could tell).

Lenetis: The tests will indeed have different performance characteristic, but I think this isn't a big problem in this case, because placing tiles in a big rectangle is more similar to a real game scenarios than placing them in a long 5000-tile line. Using road turns would complicate this function for not much real benefit (the actual tileset has only 16 tiles with roads)

Jackenmen: > because placing tiles in a big rectangle is more similar to a real game scenarios

We have different performance tests for that:
```
BenchmarkManySmallCities
BenchmarkManySmallFields
BenchmarkManySmallMonasteries
BenchmarkManyShortRoads
```

The test I'm referring to (`BenchmarkSingleExtraLongRoad`) is part of what I'd call a set of stress tests which specifically aim to see how everything scales when you make features a lot larger:
```
BenchmarkLongCity
BenchmarkLongField
BenchmarkLongMonastery
BenchmarkSingleExtraLongRoad
```

With this change, `BenchmarkSingleExtraLongRoad` is changed to something in between scalability and synthetic testing.

